PR #427, opened 2020-08-11
Only append to /etc/hosts if entry not exists

ideal:
If `sealos init` fails and runs multiple times, there will be duplicate entries in /etc/hosts .


zhangguanzhang:
sed之前判断行存在否更好

PR #448, opened 2020-09-03
remove vendor in build.sh

oldthreefeng:
remove vendor in build.sh


oldthreefeng:
#446

oldthreefeng:
need to test every subcommand  like .  when use` upx sealos ` and add `-ldflags="-s -w"` 
```
sealos init
sealos join 
sealos cert
...
```

PR #599, opened 2021-02-25
Fix the bug that ssh missing output when a command prints multiple lines and exits quickly

SataQiu:
修复当命令行程序输出多行文本并快速退出时，程序打印结果不完整的问题


oldthreefeng:
谢谢. 
有验证过么.

SataQiu:
It looks like #488 has fixed this issue.

/close

PR #1002, opened 2022-05-13
feature: add pull commands using buildah #986

yyf1986:
[SKIP CI]seaols: 一句话简短描述该PR内容
feature: add pull commands using buildah #986

### DIFF
--- a/cmd/sealos/cmd/pull.go
+++ b/cmd/sealos/cmd/pull.go
@@ -15,6 +15,7 @@
 package cmd
 
 import (
+	"github.com/containers/buildah"
 	"github.com/labring/sealos/pkg/image"
 	"github.com/spf13/cobra"
 )
@@ -37,6 +38,9 @@ func newPullCmd() *cobra.Command {
 }
 
 func init() {
+	if buildah.InitReexec() {
+		return
+	}
 	pullCmd := newPullCmd()
 	rootCmd.AddCommand(pullCmd)
 }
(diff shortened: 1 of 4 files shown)


cmd/sealos/cmd/pull.go
```diff
@@ -37,6 +38,9 @@ func newPullCmd() *cobra.Command {
 }
 
 func init() {
+	if buildah.InitReexec() {
```
cuisongliu, line 41: buildah init  can move to servic interface new func

pkg/image/service.go
```diff
@@ -35,7 +36,15 @@ func NewRegistryService() (types.RegistryService, error) {
 	if err := checkBuildah(); err == nil {
 		return binary.NewRegistryService()
 	}
-	return nil, errors.New("buildah not found in system path")
+	err := buildahPolicySync()
```
cuisongliu, line 39: if sdk and binary need this func ,we can init it at New Service func .

cmd/sealos/cmd/pull.go
fanux: main func already has call this func?

pkg/image/buildah/registry/registry.go
```diff
@@ -0,0 +1,123 @@
+/*
+Copyright 2022 191557539@qq.com.
```
fanux, line 2: 191557539@qq.com -> sealos

cmd/sealos/cmd/pull.go
yyf1986: I fix it.this func already has call.

cuisongliu (approved):

PR #1007, opened 2022-05-13
test_for_fork

gebilxs:
**# please refuse this pull request !!!**


fanux:
@gebilxs maybe you can update something else for a test. like we already support kubernetes v1.24+ but readme not update that.

PR #1015, opened 2022-05-14
feature: add push commands using buildah #986

yyf1986:
feature: add push commands using buildah #986


pkg/image/buildah/registry/push.go
```diff
@@ -0,0 +1,167 @@
+// Copyright © 2022 sealos.
```
fanux, line 1: sealos->buildah

If we copy buildah codes, we need contains it licence

cuisongliu (approved):

PR #1041, opened 2022-05-21
add create/delete/list/inspect commands using buildah #986

yyf1986:
add create/delete/list/inspect commands using buildah #986


pkg/image/buildah/cluster/create.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	bb "github.com/labring/sealos/pkg/image/buildah/cluster/buildah"
+	"github.com/labring/sealos/pkg/image/types"
+	"github.com/pkg/errors"
+)
+
+func (c *ClusterService) Create(name string, image string) (*types.ClusterManifest, error) {
+	// delete
+	err := c.Delete(name)
```
cuisongliu, line 27: same is `buildah rm` ?  if container is not exists, is pass?

fanux (approved):

pkg/image/buildah/cluster/create.go
yyf1986: I refer to the command line implementation,delete contains `buildah umount` and `build rm`

cuisongliu (approved):

Issue #1022, opened 2022-05-15
support cache http registry


fanux:
If you try to save a http repo:

sealos.hub:5000/nvidia/k8s/container-toolkit:v1.9.0-ubuntu18.04
```
root@iZj6ce0h880med3ug4nr1mZ:~/test# sealos build -t tdst -f Dockerfile .
2022-05-15 22:43:51 [INFO] srcPath is empty manifests
2022-05-15 22:43:51 [INFO] fetch manifests images: []
2022-05-15 22:43:51 [INFO] pull images [sealos.hub:5000/nvidia/k8s/container-toolkit:v1.9.0-ubuntu18.04] for platform is linux/amd64
Pulling image: sealos.hub:5000/nvidia/k8s/container-toolkit:v1.9.0-ubuntu18.04
2022-05-15 22:43:51 [WARN] https error: Get "https://sealos.hub:5000/v2/": http: server gave HTTP response to HTTPS client, sealos try to use http
2022-05-15 22:43:51 [WARN] save domain sealos.hub:5000 image sealos.hub:5000/nvidia/k8s/container-toolkit:v1.9.0-ubuntu18.04 error: get nvidia/k8s/container-toolkit tag descriptor error: unknown tag=v1.9.0-ubuntu18.04, try "sealos login" if you are using a private registry
Status: images save success
```

PR #1116, opened 2022-06-08
feature(main): support cache http registry #1022

cuisongliu:
1. default only support token auth , add basic auth to cache registry
2. proxyauth.go (func configureAuth) if basicAuth and getAuthURLs is empty append remoteURL to default authURLs
3. proxyregistry.go (func Repository) default use tokenAuth ,if  basicAuth is true  use basicAuthHandler.
4. session.go (func AuthorizeRequest) bh.creds.Basic use remoteURL  to fetch authURLs  username password

### DIFF
--- a/cmd/sealctl/cmd/registry.go
+++ b/cmd/sealctl/cmd/registry.go
@@ -40,6 +40,7 @@ var (
 	registryPullAuths        []string
 	registryPullArch         string
 	registryPullMaxPullProcs int
+	registryPullAuthBasic    bool
 )
 
 func NewRegistryImageCmd() *cobra.Command {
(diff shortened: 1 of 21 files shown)


cuisongliu:
1. default only support token auth , add basic auth to cache registry
2. proxyauth.go (func configureAuth) if basicAuth and getAuthURLs is empty append remoteURL to default authURLs
3. proxyregistry.go (func Repository) default use tokenAuth ,if  basicAuth is true  use basicAuthHandler.
4. session.go (func AuthorizeRequest) bh.creds.Basic use default authURLs

fanux (approved):

PR #1151, opened 2022-06-23
feat: support entrypoint in Dockerfile

### LINKED ISSUE #1099: Support entrypoint in Dockerfile
entrypoint is operator friendly：

```
FROM scratch
COPY . .
ENTRYPOINT ["kubectl apply -f mysql-operator.yaml"]
CMD ["kubectl apply -f mysql-instance.yaml"]
```

So that I only need to overwrite the CMD when I want to use my own mysql config.

gitccl:
[SKIP CI]sealos: resolve #1099

### DIFF
--- a/pkg/guest/guest.go
+++ b/pkg/guest/guest.go
@@ -97,15 +97,15 @@ func (d *Default) getGuestCmd(envs map[string]string, cluster *v2.Cluster, mount
 			envs = maps.MergeMap(baseEnvs, i.Env)
 		}
 		mapping := expansion.MappingFuncFor(envs)
-		if len(i.Cmd) != 0 {
-			for _, cmd := range i.Cmd {
-				command = append(command, expansion.Expand(cmd, mapping))
-			}
+
+		for _, cmd := range i.Entrypoint {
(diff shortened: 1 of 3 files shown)


pkg/guest/guest.go
```diff
@@ -97,6 +97,11 @@ func (d *Default) getGuestCmd(envs map[string]string, cluster *v2.Cluster, mount
 			envs = maps.MergeMap(baseEnvs, i.Env)
 		}
 		mapping := expansion.MappingFuncFor(envs)
+
+		for _, cmd := range i.EntryPoint {
+			command = append(command, expansion.Expand(cmd, mapping))
+		}
+
 		if len(i.Cmd) != 0 {
```
cuisongliu, line 105: Is the judgement here unnecessary?

gitccl: I think the judgetment here is unnecessary, bacause range a nil slice is valid.

cuisongliu: You can delete this if Judgement

fanux (approved):

fanux (approved):

PR #1174, opened 2022-06-24
docs: update go version

Abingcbc:
When compiling sealos in go under 1.16, the following error will raise

> build command-line-arguments: cannot load io/fs: malformed module path "io/fs": missing dot in first path element


zzjin:
We can fix it by using `+build go1.16` to both spport go version under 1.16?

fanux (approved):

Issue #1176, opened 2022-06-24
BUG: make build in Mac arm64


cuisongliu:
```
===========> Building binary sealctl for linux_arm64
CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -tags "containers_image_openpgp" -ldflags "-X github.com/labring/sealos/pkg/version.gitVersion=untagged -X github.com/labring/sealos/pkg/version.gitCommit=4333c493 -X github.com/labring/sealos/pkg/version.buildDate=2022-06-24T23:27:38+0800" -o /Users/cuisongliu/Workspaces/go/src/github.com/labring/sealos/bin/linux_arm64/sealctl github.com/labring/sealos/cmd/sealctl
===========> Building binary sealos for linux_arm64
CGO_ENABLED=1 GOOS=linux GOARCH=arm64 go build -tags "containers_image_openpgp" -ldflags "-X github.com/labring/sealos/pkg/version.gitVersion=untagged -X github.com/labring/sealos/pkg/version.gitCommit=4333c493 -X github.com/labring/sealos/pkg/version.buildDate=2022-06-24T23:27:40+0800" -o /Users/cuisongliu/Workspaces/go/src/github.com/labring/sealos/bin/linux_arm64/sealos github.com/labring/sealos/cmd/sealos
# runtime/cgo
linux_syscall.c:67:13: error: implicit declaration of function 'setresgid' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
linux_syscall.c:67:13: note: did you mean 'setregid'?
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/unistd.h:593:6: note: 'setregid' declared here
linux_syscall.c:73:13: error: implicit declaration of function 'setresuid' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
linux_syscall.c:73:13: note: did you mean 'setreuid'?
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/unistd.h:595:6: note: 'setreuid' declared here
make[1]: *** [go.build.linux_arm64.sealos] Error 2
make: *** [build] Error 2
```

SignorMercurio:
It's impossible to build Linux binaries on macOS when `CGO_ENABLED=1`, which is required by sealos.

PR #1181, opened 2022-06-25
fix: fix bug in gen.clean and use local binaries for tools

### LINKED ISSUE #1175: BUG: make gen delete files in runtime dirs
image

we need update files in `$(ROOT_PACKAGE)/pkg/types/v1beta1` only

### LINKED ISSUE #1178: Feature:  tools.mk  download files using go-get-tool to  `bin` dir
```
PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
define go-get-tool
@[ -f $(1) ] || { \
set -e ;\
TMP_DIR=$$(mktemp -d) ;\
cd $$TMP_DIR ;\
go mod init tmp ;\
echo "Downloading $(2)" ;\
GOBIN=$(PROJECT_DIR)/bin go install $(2) ;\
rm -rf $$TMP_DIR ;\
}
endef
```

In some cases, our local files are inconsistent with the file versions used by the sealos project, so a bin is planned to put these binaries to prevent version conflicts.

SignorMercurio:
- Fix #1175
- Fix #1178, using `tools` directory to store tools binaries, which will not be deleted by `make clean`
- Maybe we should close issue #1176 as it's not a bug


scripts/make-rules/tools.mk
```diff
@@ -55,7 +55,7 @@ install.ossutil:
 install.upx:
 	@wget https://github.com/upx/upx/releases/download/v3.96/upx-3.96-amd64_linux.tar.xz
```
cuisongliu, line 56: Do you want to consider multi-architecture support? For example, arm64

SignorMercurio: Thanks for reminding, I'll implement it soon.

cuisongliu (approved):

PR #1226, opened 2022-06-30
fix: merge go.build.%.sealctl/sealos

SignorMercurio:
Now sealos will be built with `CGO_ENABLED=0` too, which makes cross-platform building possible.


cuisongliu (approved):

berlinsaint:
lgtm

PR #1242, opened 2022-07-03
Dev coverage

zzjin:
Add support test coverage in pull request and push


SignorMercurio:
Would you consider adding these steps into Makefile?

zzjin:
> Would you consider adding these steps into Makefile?

yes, move ci command to makefile will be better, but we still need install required test packages in actions step?

zzjin:
Updated, use github's `codecov` app instead of self-writing.

PR #1258, opened 2022-07-06
docs: improve DEVELOPGUIDE.md

SignorMercurio:


SignorMercurio:
I think CI/unit-test will run on both push and pull request, which should be duplicated.

zzjin (approved):

zzjin:
> h push and pull request, which should be duplicated.

Since they are not the same, push means git push code event, but pull request may have other's forks to merge, so we still need to run ci at this situation.n.

PR #1277, opened 2022-07-08
Draft: try support multi tags glog

zzjin:
Try resolve `flag redefined` panic by using build tags.

**Cons:**
this way is to change lines of source code of `glog` with inelegance, any other method can be done to not change the `glog``s source code?

**Pros:**
this way can use go's build tags to build as usual. 

Do not merge now.


SignorMercurio:
Shall we close this since we no longer use `glog` now in #1300 ?

PR #1316, opened 2022-07-15
feat: support custom apiserver port

### LINKED ISSUE #1311: Question: How to change `api-server` default port at init?
我想初始化的时候，api-server使用其他的端口，不使用6443,怎么改端口

gitccl:
resolve #1311

### DIFF
--- a/pkg/runtime/kubeadm.go
+++ b/pkg/runtime/kubeadm.go
@@ -149,14 +149,18 @@ func (k *KubeadmRuntime) getVip() string {
 	return DefaultVIP
 }
 
+func (k *KubeadmRuntime) getAPIServerPort() int32 {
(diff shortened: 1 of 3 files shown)


gitccl:
[There is still a hardcode](https://github.com/labring/sealos/blob/3a8d63703c8f21c1322d265d223d8688a18eb484/pkg/auth/utils/config.go#L121)，but I don't familiar with this module, perhaps those who are responsible for this module can change it.

cuisongliu:
> I don't know whether [this code](https://github.com/labring/sealos/blob/3a8d63703c8f21c1322d265d223d8688a18eb484/pkg/auth/utils/config.go#L121) needs to be modified，I don't familiar with this module.

The code here needs to be optimized. For example, this configuration should be read directly from kubeconfig.

cuisongliu (approved):

PR #1326, opened 2022-07-16
upgrade IsIpv4 function use to system api

zsyaoo:
upgrade IsIpv4 function use to system api


zzjin:
Thank you for first contribution!
Pls add sign-off commit to ensure contribute.
@see https://github.com/labring/sealos/pull/1326/checks?check_run_id=7369426543

cuisongliu (approved):

PR #1331, opened 2022-07-17
add infra CRD spec

fanux:


controllers/infra/api/v1/infra_types.go
```diff
@@ -0,0 +1,128 @@
...
+*/
+
+type Hosts struct {
+	Roles []string `json:"hosts,omitempty"`
+	Count int      `json:"count,omitempty"`
+	// key values resources.
+	// cpu: 2
+	// memory: 4
+	// other resources like GPU
+	Resources map[string]string `json:"resources,omitempty"`
```
zzjin, line 71:
```suggestion
	Resources map[string]interface{} `json:"resources,omitempty"`
```
should resources support number|string| subarray?

zzjin (approved):

controllers/infra/api/v1/infra_types.go
fanux: +1

fanux:
```
type Quantity struct {
	// i is the quantity in int64 scaled form, if d.Dec == nil
	i int64Amount
	// d is the quantity in inf.Dec form if d.Dec != nil
	d infDecAmount
	// s is the generated value of this quantity to avoid recalculation
	s string

	// Change Format at will. See the comment for Canonicalize for
	// more details.
	Format
}
```

Kuberentes resource using Quantity struct.

PR #1336, opened 2022-07-18
feat: support test mode for lvscare

fengxsong:
In some scenario, we want that lvscare can automatically bind virtual IP address to a dummy interface, this PR does
1. check if any of realservers is at the same lvscare host, then create dummy interface and assign virtual IP to it, otherwise do nothing.
2. test mode flag is disabled by default
3. adjust a few logger level

### DIFF
--- a/staging/src/github.com/labring/lvscare/care/care.go
+++ b/staging/src/github.com/labring/lvscare/care/care.go
@@ -16,6 +16,7 @@ package care
 
 import (
 	"errors"
+	"fmt"
 	"net"
 	"os"
 	"os/signal"
@@ -27,25 +28,40 @@ import (
 	"github.com/labring/sealos/pkg/utils/logger"
 )
 
-//VsAndRsCare is
-func (care *LvsCare) VsAndRsCare() {
-	lvs := BuildLvscare()
-	//set inner lvs
-	care.lvs = lvs
-	if care.Clean {
+// VsAndRsCare is
+func (care *LvsCare) VsAndRsCare() error {
+	if care.lvs == nil {
+		care.lvs = BuildLvscare()
+	}
+
+	cleanVirtualServer := func() error {
 		logger.Info("lvscare deleteVirtualServer")
-		err := lvs.DeleteVirtualServer(care.VirtualServer, false)
+		err := care.lvs.DeleteVirtualServer(care.VirtualServer, false)
 		if err != nil {
-			logger.Info("virtualServer is not exist skip...: %v", err)
+			logger.Warn("virtualServer is not exist skip...: %v", err)
 		}
+		return nil
 	}
-	care.createVsAndRs()
+
+	if care.Clean {
+		cleanVirtualServer()
+		care.cleanupFuncs = append(care.cleanupFuncs, cleanVirtualServer)
+	}
+	if err := care.test(); err != nil {
+		return err
+	}
+	if err := care.createVsAndRs(); err != nil {
+		return err
+	}
 	if care.RunOnce {
-		return
+		return nil
 	}
+
 	t := time.NewTicker(time.Duration(care.Interval) * time.Second)
+	defer t.Stop()
 	sig := make(chan os.Signal, 1)
 	signal.Notify(sig, syscall.SIGINT, syscall.SIGTERM)
+
 	for {
 		select {
 		case <-t.C:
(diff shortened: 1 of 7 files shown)


staging/src/github.com/labring/lvscare/care/vars.go
```diff
@@ -27,12 +27,16 @@ type LvsCare struct {
...
-	Route *route.Route
+	// runtime
+	cleanupFuncs []func() error
+	lvs          Lvser
+	Route        *route.Route
 }
 
 var LVS LvsCare
+
+const dummyIfaceName = "lvscare"
```
cuisongliu, line 42: This parameter is more dynamic.

staging/src/github.com/labring/lvscare/care/care.go
```diff
@@ -27,70 +29,141 @@ import (
...
-		err := lvs.DeleteVirtualServer(care.VirtualServer, false)
+		err := care.lvs.DeleteVirtualServer(care.VirtualServer, false)
 		if err != nil {
-			logger.Info("virtualServer is not exist skip...: %v", err)
+			logger.Warn("virtualServer is not exist skip...: %v", err)
 		}
+		return err
+	}
+
+	if care.Clean {
```
cuisongliu, line 54: This parameter may not be correct. Previously, it was designed to automatically clear ipvs and route before execution. Is it of great significance to see this at present?

fengxsong: I think if we specified this parameter, lvscare should clean ipvs rules before exiting?

staging/src/github.com/labring/lvscare/care/care.go
```diff
@@ -27,70 +29,141 @@ import (
...
+		}
+	}
+	if len(errs) > 0 {
+		return errors.New(strings.Join(errs, ", "))
+	}
+	return nil
+}
+
+func (care *LvsCare) test() error {
+	if care.Test {
```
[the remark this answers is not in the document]
fengxsong: more concrete😏maybe rename it to `setupDummyIfaceOrSkip`😁

staging/src/github.com/labring/lvscare/care/care.go
cuisongliu: Yes, this parameter should indeed be like this. However, the previous code did not do this, but deleted the corresponding IPVS and routing before execution. If you want to do this, the previous functions also need to be supported.

staging/src/github.com/labring/lvscare/care/care.go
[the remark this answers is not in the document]
cuisongliu: This is a good suggestion.

staging/src/github.com/labring/lvscare/care/care.go
fengxsong: Seems `--clean` flag only useful when the last time lvscare run with flag `--run-once`. So in my understanding, this parameter is not so important as lvscare will takes care of the maintenance ipvs/route rules automatically (before exiting). 
BTW, the `previous functions` you mentioned, is this one https://github.com/labring/sealos/blob/main/cmd/sealctl/cmd/ipvs.go#L32 ?

cuisongliu: By default, IPVS and routing will be cleaned up for each execution. So this function can be used as the default. --clean's parameters can be kept now.

cuisongliu:
This PR may be changed a lot, and this function may be more important. You can take it apart and let's take a look one by one.

staging/src/github.com/labring/lvscare/care/care.go
fengxsong: left it as it is.

fanux:
@fengxsong the PR should provide a test report, I'm not sure if it will affect normal functionality

staging/src/github.com/labring/lvscare/care/vars.go
```diff
@@ -27,12 +28,31 @@ type LvsCare struct {
...
+}
+
+func (l *LvsCare) RegisterFlags(fs *pflag.FlagSet) {
+	fs.IPVar(&l.TargetIP, "ip", nil, "target ip")
+	fs.BoolVar(&l.RunOnce, "run-once", false, "is run once mode")
+	fs.StringVar(&l.VirtualServer, "vs", "", "virtual server like 10.54.0.2:6443")
+	fs.StringSliceVar(&l.RealServer, "rs", []string{}, "real server like 192.168.0.2:6443")
+	fs.StringVar(&l.Logger, "logger", "INFO", "logger level: DEBG/INFO")
+	fs.BoolVarP(&l.Clean, "clean", "C", false, "before run clean ipvs rules")
+	fs.BoolVarP(&l.Test, "test", "t", false, "use when any of real server is listening on the same host, "+
```
cuisongliu, line 49: Is it better to delete this parameter? Is it better to default the iface parameter to null?

staging/src/github.com/labring/lvscare/care/care.go
```diff
@@ -27,70 +29,144 @@ import (
...
 		}
-		logger.Info("tip: %s,vip: %s", LVS.TargetIP.String(), vIP)
-		LVS.Route = route.NewRoute(vIP, LVS.TargetIP.String())
-		return LVS.Route.SetRoute()
+		logger.Info("tip: %s,vip: %s", care.TargetIP.String(), vIP)
+		care.Route = route.NewRoute(vIP, care.TargetIP.String())
+		if err = care.Route.SetRoute(); err != nil {
+			return err
+		}
+		if !care.RunOnce {
```
cuisongliu, line 166: It doesn't seem necessary to set up this here, right?

staging/src/github.com/labring/lvscare/care/vars.go
```diff
@@ -27,12 +28,31 @@ type LvsCare struct {
...
+	Route        *route.Route
+}
+
+func (l *LvsCare) RegisterFlags(fs *pflag.FlagSet) {
+	fs.IPVar(&l.TargetIP, "ip", nil, "target ip")
+	fs.BoolVar(&l.RunOnce, "run-once", false, "is run once mode")
+	fs.StringVar(&l.VirtualServer, "vs", "", "virtual server like 10.54.0.2:6443")
+	fs.StringSliceVar(&l.RealServer, "rs", []string{}, "real server like 192.168.0.2:6443")
+	fs.StringVar(&l.Logger, "logger", "INFO", "logger level: DEBG/INFO")
+	fs.BoolVarP(&l.Clean, "clean", "C", false, "before run clean ipvs rules")
```
cuisongliu, line 48: I took a look. The parameters here can be consistent with the meaning of sealtl ipvs -C. Just clean it up.

cuisongliu:
resolve conflicts

fengxsong:
I'll make another two pull requests with those refactor and feature. so close it.

PR #1367, opened 2022-07-21
feat: auto create dummy link if needed

fengxsong:
automatic create a dummy network interface and bind virtual IP to it when any of real server is listening locally(by automatic dectection), setting this parameter to null will skip this behaviour.

### DIFF
--- a/staging/src/github.com/labring/lvscare/care/vars.go
+++ b/staging/src/github.com/labring/lvscare/care/vars.go
@@ -29,6 +29,7 @@ type LvsCare struct {
 	RunOnce       bool
 	Clean         bool
 	Interval      int32
+	IfaceName     string
 	Logger        string
 	TargetIP      net.IP
 	// runtime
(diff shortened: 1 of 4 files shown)


fengxsong:
The usage scenario is that we want to use lvscare static pod to do load balancing on the control plane. Without this function, we must maintained net link manually, like 

```
# setup
sudo ip link add dev lvscare type dummy
sudo ip addr add 10.54.0.2/32 dev lvscare
# cleanup
sudo ip link delete lvscare
```

If these rules are missing, when ipvs forwards tcp requests to the real backend on the local machine, no route to host error will occured.

staging/src/github.com/labring/lvscare/care/vars.go
```diff
@@ -42,6 +43,9 @@ func (care *LvsCare) RegisterFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&care.RunOnce, "run-once", false, "run once mode, creating ipvs rules and routes and exit")
 	fs.StringVar(&care.VirtualServer, "vs", "", "virtual server like 10.54.0.2:6443")
 	fs.StringSliceVar(&care.RealServer, "rs", []string{}, "real server like 192.168.0.2:6443")
+	fs.StringVar(&care.IfaceName, "iface", "lvscare", "name of interface to created if needed, "+
```
cuisongliu, line 46: The default is empty. If not, it may affect the function.

cuisongliu (approved):

PR #1370, opened 2022-07-21
update auth service

zzjin:
Signed-off-by: zzjin <tczzjin@gmail.com>


service/auth/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM scratch
+LABEL image="ghcr.io/labring/sealos/service/auth"
+COPY main ./main
+COPY auth.yaml auth.yaml
+CMD ["./main", "-config", "./auth.yaml"]
```
Abingcbc, line 5: Maybe we can use multi-stage dockerfile build here?
Golang build is the first stage, and here is the second and final stage

zzjin: SelfMake using docker file? or just run make file? I'm not sure which is better for global mantance.
Butirst time we can use docker file to build it, and combining outputs to cluster yaml?
@SignorMercurio Any suggestion?

Abingcbc: We can still make use of Makefile in the first stage like `CMD ["make", "build", "auth"]
This will simplify the release process into one step (directly build the image and push it to dockerhub). Otherwise, we have to prepare a golang environment first to build the binary file and then we can build the image.

I assume that the release of binary and docker images are two individual CI processes. If we release them together, I agree with current dockerfile.

SignorMercurio: Multi-stage building is a cool feature, please feel free to contribute. But at present, binary releasing and docker image publishing are within a single CI process (see .goreleaser.yml), and I'm not sure whether we should separate them because doing so needs to tamper with GoReleaser configs.

SignorMercurio: @zzjin I'm not familiar with this module, but I believe `service/auth` here produces a binary. According to the existing directory structure, should we move these into `cmd/`? From the `Dockerfile` I guess that the binary is named `main`, should we change it? Finally, I suggest moving any `Dockerfile` to `docker/` so that we can possibly `make image.build` and `make image.push` in the future.

zzjin: At this first stage, packages(currently only `auth`, but will add more soon) under `service` first build one binary, and combine this binary and config file(s) with many other requirements into a full cluster image.

So we should push the cluster image to hub but not the binary docker file itself.
In the other words, we should not push(or publish) binary's Dockerfile.

service/auth/deploy/manifests/auth.yaml
```diff
@@ -0,0 +1,56 @@
...
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app: service-auth
+    spec:
+      containers:
+        - name: service-auth
+          image: service-auth:latest
```
Abingcbc, line 56: So we will put this image into cluster image?

Abingcbc: Cluster image seems to only be responsible for storeing claims of cluster, like a `Deployment` yaml. But the individual image of each container is managed by the registry.

zzjin: yes，what we need a two step ci，one for build service－auth，then the later one build cluster image.

service/auth/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM scratch
+LABEL image="ghcr.io/labring/sealos/service/auth"
```
cuisongliu, line 2: Does the label change the tag? It conflicts with the default rootfs.

service/auth/deploy/Kubefile
```diff
@@ -0,0 +1,3 @@
+FROM scratch
+COPY . .
```
cuisongliu, line 2: Try not to copy it like this, and the default Kubefile will be overwritten.

service/auth/Dockerfile
cuisongliu: You can directly make a cluster image, start a registry locally, and push the mirror to the mirror repository. The final product is cluster mirroring.

fanux (approved):

PR #1376, opened 2022-07-21
remove duplicate codes

fengxsong:

remove duplicate code in ClusterService

### DIFF
--- a/pkg/image/binary/cluster.go
+++ b/pkg/image/binary/cluster.go
@@ -32,39 +32,28 @@ import (
 type ClusterService struct {
 }
 
-func (d *ClusterService) Create(name string, image string) (*types.ClusterManifest, error) {
-	data := exec.BashEval("buildah containers --json")
-	infos, err := listContainer(data)
-	if err != nil {
+func (s *ClusterService) Create(name string, image string) (*types.ClusterManifest, error) {
+	if err := s.Delete(name); err != nil {
 		return nil, err
 	}
 
-	for _, info := range infos {
-		if info.Containername == name {
-			cmd := fmt.Sprintf("buildah unmount %s && buildah rm  %s", info.Containername, info.Containername)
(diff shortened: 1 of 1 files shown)


SignorMercurio:
Please sync the code with `main` as there're previous commits that have made CI failed.

pkg/image/binary/cluster.go
```diff
@@ -32,39 +32,28 @@ import (
...
-	infos, err := listContainer(data)
-	if err != nil {
+func (s *ClusterService) Create(name string, image string) (*types.ClusterManifest, error) {
+	if err := s.Delete(name); err != nil {
 		return nil, err
 	}
 
-	for _, info := range infos {
-		if info.Containername == name {
-			cmd := fmt.Sprintf("buildah unmount %s && buildah rm  %s", info.Containername, info.Containername)
```
zhangguanzhang, line 44: why removed the umount logic

fengxsong: duplicate code block, you should check the whole file.

pkg/image/binary/cluster.go
```diff
@@ -32,39 +32,28 @@ import (
...
-		}
-	}
-
-	cmd := fmt.Sprintf(" buildah from --pull=never --name %s %s && buildah mount %s ", name, image, name)
-	err = exec.Cmd("bash", "-c", cmd)
+	cmd := fmt.Sprintf("buildah from --pull=never --name %s %s && buildah mount %s ", name, image, name)
+	err := exec.Cmd("bash", "-c", cmd)
 	if err != nil {
 		return nil, err
 	}
```
zzjin, line 44:
```suggestion
	if err := exec.Cmd("bash", "-c", cmd); err != nil {
		return nil, err
	}
```
this can avoid `err` reassign.

fanux (approved):

PR #1403, opened 2022-07-26
support sudo #690

yyf1986:
useradd -m -s /bin/bash -b /home/sealos
sealos ALL=(ALL:ALL) NOPASSWD:ALL

sudo sealos run labring/kubernetes:v1.24.0 labring/calico:v3.22.1 --masters 192.168.175.149 -p sealos -u sealos


fanux:
This PR has tooo many conflict, and we may needs rethink how to support sudo.

Issue #1364, opened 2022-07-21
BUG: 文档在语言为简体中文时, 乱入了一节英文内容


FsXI:
网站地址: https://www.sealos.io/zh-Hans/docs/cli/build

截图:
![image](https://user-images.githubusercontent.com/40119641/180142363-d05e2443-d5ce-4640-b846-95e574e404c5.png)

zzjin:
There has some documents not been translated to `zh-Hans`.
PRs are welcome~~~ 
Just click "编辑此页" can edit&translate it online.

fanux:
@FsXI 可以在 github.com/fanux/sealos-site 这个仓库中帮我们修复

SignorMercurio:
Completed via #1416

PR #1416, opened 2022-07-27
fix docs that have english

willzhang:
fix docs that have english， the issue: https://github.com/labring/sealos/issues/1364


fanux:
git commit -s --amend

cuisongliu (approved):

PR #1426, opened 2022-07-28
add-registry-docs

willzhang:
Signed-off-by: will <willzhmic@outlook.com>

add-registry-docs


docs/4.0/docs/examples/use-private-registry.md
```diff
@@ -0,0 +1,69 @@
+# how to use sealos private registry
```
SignorMercurio, line 1:
```suggestion
# How to use sealos private registry
```

Please use upper case letters properly (e.g. First letter of sentences, topics, etc.)

docs/4.0/docs/examples/use-private-registry.md
```diff
@@ -0,0 +1,69 @@
+# how to use sealos private registry
+
+## where it is
+
+the sealos private registry running in the first nodes of the cluster,where you create your cluster, you can use the follow comand check it:
+
+```
```
SignorMercurio, line 7: Please specify a language for each code blocks. For example:

<pre>
```shell
$ nerdctl ps
```
</pre>

But there's no need to do this when it's not related to a language, like:

<pre>
```
/var/lib/sealos/data/default/rootfs/registry/
```
</pre>

docs/4.0/docs/examples/use-private-registry.md
```diff
@@ -0,0 +1,69 @@
...
+
+the registry images data store in  follow directory:
+
+```
+/var/lib/sealos/data/default/rootfs/registry/
+```
+
+## login to registry
+
+sealos private registry run with http and `--net host ` option，you should config insecure-registry at local and connect with first node ipaddress，example of docker client:
```
SignorMercurio, line 21:
```suggestion
sealos private registry runs with HTTP and `--net host` option. You should configure `insecure-registries` at local and connect with the first node's IP address. Example of docker client:
```

Please make sure all the punctuations are English punctuations, not the Chinese ones.

SignorMercurio (changes requested):

SignorMercurio:
Why is this PR closed?

PR #1428, opened 2022-07-28
fix(auth): connect to casdoor

Abingcbc:
Fix:
1. Each time restarting service-auth, load the existing Casdoor init data instead of creating a new one.
2. Redirect URL to Casdoor (ssoEndpoint in config.yaml)
3. Wrap all errors with `errors.Wrap`
4. Parsing OAuth token error. 
    The root cause is that Casdoor raised a breaking change. In PR https://github.com/casdoor/casdoor/pull/894, `PublicKey` is renamed to `Certificate`. Therefore, the jwt publickey we created is ignored by Casdoor.

### DIFF
--- a/pkg/auth/sso/casdoor.go
+++ b/pkg/auth/sso/casdoor.go
@@ -18,31 +18,33 @@ import (
 	"context"
 	"encoding/json"
 
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"github.com/pkg/errors"
+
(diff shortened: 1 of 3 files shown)


zzjin (approved):

Abingcbc:
@zzjin @fanux PTAL

zzjin:
Pls fix the linter error, and then wen can merge it.

zzjin (approved):

PR #1433, opened 2022-07-29
fix chart parse image failed.

yyf1986:
fix_chart_1431
default like:
```
- name: ***
  image: ***
```
support chart image like 
```
- image: ***
  name: ***
```


pkg/buildimage/chart.go
```diff
@@ -113,12 +113,18 @@ func (c Chart) GetImages() ([]string, error) {
...
 		return shaReg.ReplaceAllString(a, "")
 	}
+	// add del -image: and image:
+	delImageReg := regexp.MustCompile(`-{0,1}image:`)
+	delImage := func(str string) string {
+		return delImageReg.ReplaceAllString(str, "")
+	}
 
 	for _, v := range content {
 		if delLF(v) != "" { // Text has content
```
fanux, line 123: if delLF(v) == "" { continue } ...

pkg/buildimage/chart.go
```diff
@@ -113,12 +113,18 @@ func (c Chart) GetImages() ([]string, error) {
 	delImageSha := func(a string) string {
 		return shaReg.ReplaceAllString(a, "")
 	}
+	// add del -image: and image:
+	delImageReg := regexp.MustCompile(`-{0,1}image:`)
+	delImage := func(str string) string {
+		return delImageReg.ReplaceAllString(str, "")
+	}
 
 	for _, v := range content {
```
fanux, line 122: Use a parse function, and add some unitest for this function:

func parseImages(content) []string {}

pkg/buildimage/chart.go
yyf1986: what's meaning?

pkg/buildimage/chart.go
yyf1986: Need to write a unit test?

fanux: yes, unit test for this function is important

pkg/buildimage/chart.go
fanux, line 123:
```suggestion
		if delLF(v) == "" { 
		    continue
		}
		for _,s := range ....
		```

fanux:
@yyf1986 rebase

yyf1986:
resubmit

PR #1438, opened 2022-07-31
fix chart parse image  and add unit test

yyf1986:
Signed-off-by: yyf1986 <191557539@qq.com>

fix_chart_1431
default like:
```
- name: ***
  image: ***
```

support chart image like:
```
- image: ***
  name: ***
```

### DIFF
--- a/pkg/buildimage/chart_test.go
+++ b/pkg/buildimage/chart_test.go
@@ -1,0 +1,27 @@
+// Copyright © 2022 sealos.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


pkg/buildimage/chart_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+package buildimage
+
+import (
+	"testing"
+)
+
+func TestGetImage(t *testing.T) {
+	ccc := Chart{}
+	aaa := ccc.getImage("- image: quay.io/cilium/operator-generic:v1.11.0@sha256:b522279577d0d5f1ad7cadaacb7321d1b172d8ae8c8bc816e503c897b420cfe3")
```
cuisongliu, line 23: Use the standard test case and take a look at the test case of the apply module.

pkg/buildimage/chart_test.go
```diff
@@ -0,0 +1,27 @@
...
+// limitations under the License.
+
+package buildimage
+
+import (
+	"testing"
+)
+
+func TestGetImage(t *testing.T) {
+	ccc := Chart{}
```
fanux, line 22: bbb aaa ccc? This is not a good naming convention

cuisongliu (approved):

PR #1439, opened 2022-07-31
fix(auth): get ca cert from kubeconfig

Abingcbc:
Get CA Cert from kubeconfig instead of :
1. From `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` for in cluster mode
2. From config.CaCert for out of cluster mode


Abingcbc:
@cuisongliu PTAL

cuisongliu (approved):

PR #1444, opened 2022-08-01
add create/get aws instance success

fanux:


controllers/infra/drivers/reconcile.go
```diff
@@ -28,7 +28,8 @@ func (a *Applier) ReconcileInstance(infra *v1.Infra, driver Driver) (*v2.Cluster
 	// get current hosts
 	tag := infra.GetInstancesTag()
 	var err error
-	a.currentHosts, err = driver.GetInstancesByLabel(common.InfraInstancesLabel, tag)
+	hosts, err := driver.GetInstancesByLabel(common.InfraInstancesLabel, tag, infra)
+	a.currentHosts = append(a.currentHosts, *hosts)
 	if err != nil {
 		return nil, fmt.Errorf("failed to query instances: %v", err)
 	}
```
zzjin, line 35:
```suggestion
	hosts, err := driver.GetInstancesByLabel(common.InfraInstancesLabel, tag, infra)
	if err != nil {
		return nil, fmt.Errorf("failed to query instances: %v", err)
	}
	a.currentHosts = append(a.currentHosts, *hosts)
```

cuisongliu:
Https://kubebuilder.io/reference/using-finalizers.html Do you want to optimize this? I wrote a sdk before that can be used directly.  https://github.com/labring/endpoints-operator/blob/main/library/controller/controller.go

controllers/infra/drivers/reconcile.go
fanux: done

fanux:
> [Https://kubebuilder.io/reference/using-finalizers.html](Https://kubebuilder.io/reference/using-finalizers.html) Do you want to optimize this? I wrote a sdk before that can be used directly. https://github.com/labring/endpoints-operator/blob/main/library/controller/controller.go

yes I will

cuisongliu (approved):

PR #1446, opened 2022-08-01
add delete aws instance

fanux:

aws only has stop instance interface, it will auto clean the stoped instance.


controllers/infra/drivers/aws/stop_instance.go
```diff
@@ -0,0 +1,81 @@
...
+	if errors.As(err, &apiErr) && apiErr.ErrorCode() == "DryRunOperation" {
+		fmt.Println("User has permission to stop instances.")
+		input.DryRun = aws.Bool(false)
+		return api.StopInstances(c, input)
+	}
+
+	return resp, err
+}
+
+func (d Driver) deleteInstances(hosts *v1.Hosts) error {
```
SignorMercurio, line 63: I think we'd better add some comments here because `delete` instances are easy to obfuscate with `stop` instances here.

SignorMercurio (approved):

controllers/infra/drivers/aws/stop_instance.go
fanux: aws will auto delete the stoped instance, but for this interface we want to delete the instance not stop it.

PR #1489, opened 2022-08-07
Set aws infra name & index tag. (#1462)

Ficus-f:
Set aws infra name & index tag. (#1462)

Signed-off-by: Ficus <ficusf@qq.com>


fanux:
@Ficus-f   
```
controllers/infra/drivers/aws/create_instance.go:22: File is not `goimports`-ed (goimports)
  	"github.com/labring/sealos/controllers/infra/common"
```

fanux (approved):

Issue #1490, opened 2022-08-07
Feature: reconcile aws infra count


Ficus-f:

HURUIZHE:
Completely

PR #1491, opened 2022-08-07
Reconcile aws infra count. (#1490)

Ficus-f:
Feature: reconcile aws infra count. (#1490)

### DIFF
--- a/controllers/infra/drivers/aws/get_instance.go
+++ b/controllers/infra/drivers/aws/get_instance.go
@@ -126,6 +126,9 @@ func (d Driver) getInstancesByLabel(key string, value string, infra *v1.Infra) (
 
 	for _, r := range result.Reservations {
 		for _, i := range r.Instances {
+			if i.State.Name == types.InstanceStateNameShuttingDown || i.State.Name == types.InstanceStateNameTerminated || i.State.Name == types.InstanceStateNameStopping || i.State.Name == types.InstanceStateNameStopped {
+				continue
+			}
 			hosts.Count++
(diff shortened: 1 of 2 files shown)


fanux:
you need `git rebase upstream/main`

controllers/infra/drivers/aws/get_instance.go
```diff
@@ -126,6 +126,9 @@ func (d Driver) getInstancesByLabel(key string, value string, infra *v1.Infra) (
 
 	for _, r := range result.Reservations {
 		for _, i := range r.Instances {
+			if i.State.Name == types.InstanceStateNameShuttingDown || i.State.Name == types.InstanceStateNameTerminated || i.State.Name == types.InstanceStateNameStopping || i.State.Name == types.InstanceStateNameStopped {
```
fanux, line 129: Instances in the stopped or stopping state cannot be ignored. Shutting down and terminated will auto clean by aws

Ficus-f: Terminated aws will cause an error if we want to get its PrivateIpAddress when the aws terminates but is not cleaned.

controllers/infra/drivers/aws/get_instance.go
```diff
@@ -126,6 +126,9 @@ func (d Driver) getInstancesByLabel(key string, value string, infra *v1.Infra) (
...
+				continue
+			}
```
fanux, line 131: if stop {
   start 
}

fanux (approved):

PR #1496, opened 2022-08-08
fix: only snat packets marked

fengxsong:

prevent doing snat for packets those didn't mark.


fengxsong:
Incompatible with cilium proxyless mode, I'am working on it.

fanux (approved):

PR #1501, opened 2022-08-09
fix: incompatible with cilium

fengxsong:

1. ruler will create a goroutine to ensure iptables rules every 1 minute incase some external program flush iptables rules;
2. cilium CNI will create rules in `filter` table, non-matched marked packets will be dropped, so lvscare must ACCEPT those packets at the very first chain.

### DIFF
--- a/staging/src/github.com/labring/lvscare/care/iptables.go
+++ b/staging/src/github.com/labring/lvscare/care/iptables.go
@@ -17,7 +17,9 @@ package care
 import (
 	"errors"
 	"fmt"
+	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	utilsysctl "k8s.io/component-helpers/node/util/sysctl"
 	proxyipvs "k8s.io/kubernetes/pkg/proxy/ipvs"
 	utilipset "k8s.io/kubernetes/pkg/util/ipset"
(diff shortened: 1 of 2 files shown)


staging/src/github.com/labring/lvscare/care/iptables.go
```diff
@@ -200,6 +202,33 @@ func (impl *iptablesImpl) ensureIpset() error {
...
+	position utiliptables.RulePosition
+	table    utiliptables.Table
+	chain    utiliptables.Chain
+	args     []string
+}
+
+func (impl *iptablesImpl) extraChainRules() []iptablesRule {
+	rules := make([]iptablesRule, 0)
+	if exists, _ := impl.iptables.ChainExists(utiliptables.TableFilter, "CILIUM_OUTPUT"); exists {
+		// another work around is set `prependIptablesChains=false` in cilium agent's config
```
fengxsong, line 215: OR maybe this is a common solution? @cuisongliu

fanux: kube-proxy has extraIP config, don't have extraChainRules config, It's embarrassing, it's constantly being cleaned and rebuilt.

fanux (approved):

PR #1553, opened 2022-08-17
fix(main): add docker-shim support

cuisongliu:

### DIFF
--- a/staging/src/github.com/labring/image-cri-shim/pkg/cri/cri_unix.go
+++ b/staging/src/github.com/labring/image-cri-shim/pkg/cri/cri_unix.go
@@ -30,7 +30,8 @@ const (
 	CRISocketCRIO = "unix:///var/run/crio/crio.sock"
 	// CRISocketDocker is the cri-dockerd CRI endpoint
 	CRISocketDocker = "unix:///var/run/cri-dockerd.sock"
-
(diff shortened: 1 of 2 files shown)


staging/src/github.com/labring/image-cri-shim/pkg/cri/cri_unix.go
```diff
@@ -30,7 +30,8 @@ const (
 	CRISocketCRIO = "unix:///var/run/crio/crio.sock"
 	// CRISocketDocker is the cri-dockerd CRI endpoint
 	CRISocketDocker = "unix:///var/run/cri-dockerd.sock"
-
+	// CRISocketDockerLower is the cri-dockerd CRI endpoint
+	CRISocketDockerLower = "unix:///var/run/dockershim.sock"
```
fanux, line 34: Can it be used directly?

fanux (approved):

staging/src/github.com/labring/image-cri-shim/pkg/cri/cri_unix.go
cuisongliu: > Can it be used directly?

I haven't tested this PR yet. I need to test it to directly make sealos compatible with high and low versions of docker. We just need to make different cluster images to support different dockers.

PR #1574, opened 2022-08-18
fix: unexpected config merge

fengxsong:

Merge config into file in specified application image. defaults is overwriting all configs into all images(works but not necessary), when `Config.spec.strategy` is `merge`, modified files will be overwritten with null values again in later for-loop processing.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -52,15 +52,18 @@ type Interface interface {
 
 type Dumper struct {
 	Configs  []v1beta1.Config
+	name     string
 	RootPath string
 }
 
-func NewConfiguration(rootPath string, configs []v1beta1.Config) Interface {
+func NewConfiguration(name, rootPath string, configs []v1beta1.Config) Interface {
 	return &Dumper{
 		RootPath: rootPath,
+		name:     name,
 		Configs:  configs,
 	}
 }
+
 func NewDefaultConfiguration(clusterName string) Interface {
 	return &Dumper{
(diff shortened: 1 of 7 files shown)


pkg/config/config.go
```diff
@@ -52,15 +52,18 @@ type Interface interface {
 
 type Dumper struct {
 	Configs  []v1beta1.Config
+	name     string
 	RootPath string
 }
 
-func NewConfiguration(rootPath string, configs []v1beta1.Config) Interface {
+func NewConfiguration(name, rootPath string, configs []v1beta1.Config) Interface {
```
fengxsong, line 59: convert this exported function to builder mode like using `OptionFunc` to avoid modification frequently?

pkg/config/config.go
```diff
@@ -129,7 +124,7 @@ func (c *Dumper) WriteFiles() (err error) {
 func getAppendOrInsertConfigData(path string, data []byte, insert bool) ([]byte, error) {
 	var configs [][]byte
 	context, err := os.ReadFile(filepath.Clean(path))
-	if err != nil {
+	if err != nil && !os.IsNotExist(err) {
```
fengxsong, line 127: ignore `NotExist` error so Dumper can create a new file later.

fanux: seems if os.IsNotExist(err) {create file } is better?

fengxsong: not need to create file manually here, file will be created later in `file.WriteFile`

fanux (approved):

PR #1581, opened 2022-08-19
refactor: optimize the creation of terminal ingress

gitccl:
Signed-off-by: gitccl <chenchuanle6@gmail.com>

### DIFF
--- a/controllers/terminal/controllers/ingress.go
+++ b/controllers/terminal/controllers/ingress.go
@@ -17,97 +17,87 @@ limitations under the License.
 package controllers
 
 import (
-	"bytes"
-	"text/template"
-
 	certv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	cmmeta "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
 	networkingv1 "k8s.io/api/networking/v1"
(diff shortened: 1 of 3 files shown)


controllers/terminal/controllers/ingress.go
```diff
@@ -17,97 +17,87 @@ limitations under the License.
...
-	data, err := parseTemplate(CertTmpl, terminal)
-	if err != nil {
-		return nil, err
+func createIngress(terminal *terminalv1.Terminal) *networkingv1.Ingress {
+	objectMeta := metav1.ObjectMeta{
+		Name:      terminal.Name,
+		Namespace: terminal.Namespace,
+		Annotations: map[string]string{
+			"cert-manager.io/issuer":                     ClusterIssuerName,
+			"kubernetes.io/ingress.class":                "nginx",
```
fanux, line 58: what if user using other ingress controller?

fanux (approved):

controllers/terminal/controllers/ingress.go
zzjin: service-auth/casdoor also use ingress-nginx, It would be better to support multiple ignress providers but as for now ingress-nginx will best match.

PR #1646, opened 2022-08-28
translate en docs

fanux:


zzjin:
May be this could split into two pr, One docs change , one controller change.

zzjin (approved):

PR #1664, opened 2022-08-30
feat(auth): local cdn

Abingcbc:


zzjin:
Using docker layer build will do better.

Sample:
```dockerfile
FROM alpine/git as builder

WORKDIR /home
# Copy the Go Modules manifests
RUN git clone https://github.com/casbin/static.git

FROM apline/nginx
...
...
...

```

zzjin (approved):

PR #1699, opened 2022-09-04
optimize go step action

zzjin:
Signed-off-by: zzjin <tczzjin@gmail.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,12 +28,15 @@ jobs:
         with:
           repository: ${{ github.event.pull_request.head.repo.full_name }}
           ref: ${{ github.head_ref }}
-      - name: Setup Go
-        uses: actions/setup-go@v3
+
+      - name: Setup Golang with cache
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -111,24 +113,22 @@ jobs:
...
         uses: actions/checkout@v3
-      - name: Setup Go
-        uses: actions/setup-go@v3
+        with:
+          fetch-depth: 0
+      - name: Setup Golang with cache
+        uses: magnetikonline/action-golang-cache@v3
         with:
           go-version: ${{ env.GO_VERSION }}
-
```
SignorMercurio, line 118: Why there are many empty lines deleted? I think we should have an empty line between every two steps.

zzjin: there has no common usage, so jobs are new line seperated,steps will be no new line, or all steps have new line is
 better?

cuisongliu (approved):

PR #1744, opened 2022-09-09
fix: return error if any error occured

fengxsong:

When any error occurs during the save image process, the build process will be interrupted because the resulting artifact will be useless in an offline environment.


fengxsong:
reopen

fengxsong:
reopen

PR #1758, opened 2022-09-14
feature(main): add user-api

cuisongliu:


zzjin (approved):

fanux:
Using label filter to query user groups and other resources, add label when create resources. So we don't need an extra HTTP server. Frontend only need kubeconfig.

PR #1799, opened 2022-09-25
feat: generate kubeconfig by user controller

Abingcbc:

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -15,6 +15,8 @@
 package auth
 
 import (
+	"time"
+
 	"github.com/pkg/errors"
 	"golang.org/x/oauth2"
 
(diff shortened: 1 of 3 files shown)


pkg/auth/auth.go
```diff
@@ -51,15 +51,35 @@ func GetUserInfo(accessToken string) (*sso.User, error) {
...
 		return "", errors.Wrap(err, "Get user info failed")
 	}
 
-	kubeConfig, err := utils.GenerateKubeConfig(user.ID)
+	err = utils.CreateKubeConfig(user.ID)
+	if err != nil {
+		return "", errors.Wrap(err, "Create kube config failed")
+	}
+	// Wait for user controller to write kubeconfig into status
+	for i := 0; i < 10; i++ {
```
zzjin, line 65: Just retry on by one or add some sleep? eg: 50ms?

Abingcbc: done

zzjin (approved):

PR #1808, opened 2022-09-27
CREATE SSH pair key when creating instance

HURUIZHE:
Signed-off-by: HURUIZHE <553834398@qq.com>


fanux (approved):

controllers/infra/drivers/aws/create_instance.go
```diff
@@ -129,7 +142,15 @@ func (d Driver) createInstances(hosts *v1.Hosts, infra *v1.Infra) error {
...
 		},
 	)
+	if infra.Spec.SSH.PkName == "" {
+		if err := d.createKeyPair(infra); err != nil {
+			return err
+		}
+	}
 	keyName := infra.Spec.SSH.PkName
+	//todo use ami to search root device name
+	rootDeviceName := "/dev/xvda"
```
fanux, line 152: Seems we can add this config data in `infra.yaml`~

HURUIZHE: rootDeviceName associated to ami.
infra.yaml already had ami(image).
todo: use ami to search root device name by calling API or self-built map.

HURUIZHE: no necessary change in yaml

PR #1840, opened 2022-10-08
fix: Hydration failed

LeezQ:
fix hydration error

### DIFF
--- a/desktop/frontend/src/interfaces/session.ts
+++ b/desktop/frontend/src/interfaces/session.ts
@@ -11,17 +11,12 @@ export type UserInfo = {
   readonly avatar: string;
 };
 
-// TODO: 这里的类型定义有问题，需要修复
-export type KubeConfig = {
-  readonly id: string;
-  readonly name: string;
-  readonly avatar: string;
-};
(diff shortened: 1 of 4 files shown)


desktop/frontend/src/interfaces/session.ts
```diff
@@ -19,7 +19,7 @@ export type KubeConfig = {
 };
 
 export type Session = {
-  token: OAuthToken;
```
zzjin, line 22: `OAuthToken` has 4 elements that cannot be string.

desktop/frontend/src/interfaces/session.ts
```diff
@@ -19,9 +19,9 @@ export type KubeConfig = {
...
+  token?: OAuthToken;
   user: UserInfo;
-  kubeconfig: string;
+  kubeconfig: KubeConfig;
```
zzjin, line 24: Currently, `kubeconfig` is raw string

LeezQ: changed

zzjin (approved):

PR #1889, opened 2022-10-15
improve:make SetClusterRunArgs conform to 'return fast'

NTH19:
Signed-off-by: NTH19 <3310288189@qq.com>

This pr just reduces the nesting levels of if in func SetClusterRunArgs

### DIFF
--- a/pkg/apply/run.go
+++ b/pkg/apply/run.go
@@ -154,21 +154,23 @@ func (r *ClusterArgs) SetClusterRunArgs(imageList []string, args *RunArgs) error
 	r.cluster.Spec.Image = append(r.cluster.Spec.Image, imageList...)
 
 	// set host when cluster is not yet initialized
-	if r.cluster.CreationTimestamp.IsZero() {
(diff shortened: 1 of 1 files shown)


pkg/apply/run.go
```diff
@@ -154,21 +154,23 @@ func (r *ClusterArgs) SetClusterRunArgs(imageList []string, args *RunArgs) error
...
-				r.setHostWithIpsPort(masters, []string{v2.MASTER, string(v2.AMD64)})
-			}
-			if len(nodes) != 0 {
-				r.setHostWithIpsPort(nodes, []string{v2.NODE, string(v2.AMD64)})
-			}
-			r.cluster.Spec.Hosts = r.hosts
-		} else {
-			return fmt.Errorf("master ip(s) must specified")
+	if !r.cluster.CreationTimestamp.IsZero() {
+		return nil
```
zzjin, line 158: If not zero, there is no output logs, so do we need add one log before return?

fanux (approved):

PR #1927, opened 2022-10-21
fix: get host arch using SSH for apply/run, apply/reset and apply/scale

### LINKED ISSUE #1787: BUG: generate Clusterfile on arm host role is amd64
```
root@ecs-3e5b-0005:~# cat .sealos/default/Clusterfile
apiVersion: apps.sealos.io/v1beta1
kind: Cluster
metadata:
  creationTimestamp: "2022-09-22T07:11:26Z"
  name: default
spec:
  hosts:
  - ips:
    - 192.168.0.165:22
    - 192.168.0.82:22
    - 192.168.0.188:22
    roles:
    - master
    - amd64
```

SignorMercurio:
Should fix #1787.

I'm not sure if these changes cover all the related cases and whether I'm using SSH correctly though.

@cuisongliu PTAL.

### DIFF
--- a/pkg/apply/run.go
+++ b/pkg/apply/run.go
@@ -206,12 +206,12 @@ func (r *ClusterArgs) SetClusterResetArgs(args *ResetArgs) error {
 		masters := stringsutil.SplitRemoveEmpty(args.Cluster.Masters, ",")
 		nodes := stringsutil.SplitRemoveEmpty(args.Cluster.Nodes, ",")
 		r.hosts = []v2.Host{}
-		if len(masters) != 0 {
-			r.setHostWithIpsPort(masters, []string{v2.MASTER, string(v2.AMD64)})
-		}
-		if len(nodes) != 0 {
-			r.setHostWithIpsPort(nodes, []string{v2.NODE, string(v2.AMD64)})
-		}
+
+		clusterSSH := r.cluster.GetSSH()
+		sshClient := ssh.NewSSHClient(&clusterSSH, true)
+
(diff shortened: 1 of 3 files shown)


pkg/apply/run.go
```diff
@@ -162,12 +163,12 @@ func (r *ClusterArgs) SetClusterRunArgs(imageList []string, args *RunArgs) error
...
-			r.setHostWithIpsPort(masters, []string{v2.MASTER, string(v2.AMD64)})
-		}
-		if len(nodes) != 0 {
-			r.setHostWithIpsPort(nodes, []string{v2.NODE, string(v2.AMD64)})
-		}
+
+		clusterSSH := r.cluster.GetSSH()
+		sshClient := ssh.NewSSHClient(&clusterSSH, true)
+
+		r.setHostWithIpsPort(masters, []string{v2.MASTER, GetHostArch(sshClient, masters[0])})
```
cuisongliu, line 170: If there is something wrong with different nodes and different architectures, I feel traversed, right?

SignorMercurio: Sorry, I do not quite understand what you mean. Does "different nodes and different architectures" means that we have masters with different arch or nodes with different arch? I think we currently only support masters/nodes with the same arch.

cuisongliu: > Sorry, I do not quite understand what you mean. Does "different nodes and different architectures" means that we have masters with different arch or nodes with different arch? I think we currently only support masters/nodes with the same arch.

At present, it is like this, but in the future, different nodes should be satisfied with different architectures.

fanux: We can handle the mixed architecture case in a single new feature

fanux (approved):

Issue #1756, opened 2022-09-14
BUG:  sealos run using buildah sdk ,sealos run/create error from single platform image


cuisongliu:
at linux arm 
using buildah sdk 
```
root@sealos00:~# sealos run docker.io/labring/coredns:v0.0.1  -f
2022-09-14T14:19:23 info start to install app in this cluster
are you sure to override these app? Yes [y/yes], No [n/no] : y
99f1c97bcb67716f4e00e18900afda31fb33e41cfc4fedbd6a09dbe942b96c06
2022-09-14T14:19:24 error Applied to cluster error: create from fail: docker.io/labring/coredns:v0.0.1: image not known
2022-09-14T14:19:24 info
      ___           ___           ___           ___       ___           ___
     /\  \         /\  \         /\  \         /\__\     /\  \         /\  \
    /::\  \       /::\  \       /::\  \       /:/  /    /::\  \       /::\  \
   /:/\ \  \     /:/\:\  \     /:/\:\  \     /:/  /    /:/\:\  \     /:/\ \  \
  _\:\~\ \  \   /::\~\:\  \   /::\~\:\  \   /:/  /    /:/  \:\  \   _\:\~\ \  \
 /\ \:\ \ \__\ /:/\:\ \:\__\ /:/\:\ \:\__\ /:/__/    /:/__/ \:\__\ /\ \:\ \ \__\
 \:\ \:\ \/__/ \:\~\:\ \/__/ \/__\:\/:/  / \:\  \    \:\  \ /:/  / \:\ \:\ \/__/
  \:\ \:\__\    \:\ \:\__\        \::/  /   \:\  \    \:\  /:/  /   \:\ \:\__\
   \:\/:/  /     \:\ \/__/        /:/  /     \:\  \    \:\/:/  /     \:\/:/  /
    \::/  /       \:\__\         /:/  /       \:\__\    \::/  /       \::/  /
     \/__/         \/__/         \/__/         \/__/     \/__/         \/__/

                  Website :https://www.sealos.io/
                  Address :github.com/labring/sealos
		BuildVersion: untagged-382ff8d9


```

add buildah to usr/bin

```
root@sealos00:~# sealos run docker.io/labring/coredns:v0.0.1  -f
2022-09-14T14:21:10 info start to install app in this cluster
are you sure to override these app? Yes [y/yes], No [n/no] : y
99f1c97bcb67716f4e00e18900afda31fb33e41cfc4fedbd6a09dbe942b96c06
[...]

PR #1968, opened 2022-10-27
feature(main): fix already installed buildah

cuisongliu:

### DIFF
--- a/pkg/image/buildah.go
+++ b/pkg/image/buildah.go
@@ -17,7 +17,12 @@ limitations under the License.
 package image
 
 import (
+	"sync"
+
+	"github.com/labring/sealos/pkg/utils/logger"
+
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/json"
 
 	"github.com/labring/sealos/pkg/utils/exec"
 	fileutil "github.com/labring/sealos/pkg/utils/file"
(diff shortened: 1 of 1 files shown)


pkg/image/buildah.go
```diff
@@ -37,9 +42,43 @@ func initBuildah() (bool, error) {
...
+		versionJSON, _ := exec.RunBashCmd("buildah version")
+		var bVersion BuildahVersion
+		_ = json.Unmarshal([]byte(versionJSON), &bVersion)
+		logger.Debug("buildah version: %+v", bVersion)
+		onceBuildahVersion = bVersion
+	})
+	if onceBuildahVersion.GitCommit == "bc5080cc" {
+		return true, nil
+	}
+	return false, errors.New("system buildah is not sealos buildah,please uninstall buildah retry it")
```
fanux, line 65: What means "system buildah is not sealos buildah...."

fanux:
What can cause problems if buildah is already installed？

fanux:
What can cause problems if buildah is already installed？

cuisongliu:
> What can cause problems if buildah is already installed？

It may cause the sealos build to be unsuccessful, which is not the same version as the buildah we use, and the parameters are different.

cuisongliu:
fix https://github.com/labring/sealos/issues/1756 , delete buildah check

pkg/image/buildah.go
zzjin, line 65:
```suggestion
	return false, errors.New("System's buildah is not match sealos buildah requirements, please uninstall system's buildah, and retry")
```

zzjin (approved):

pkg/image/buildah.go
```diff
@@ -37,9 +42,43 @@ func initBuildah() (bool, error) {
...
+		versionJSON, _ := exec.RunBashCmd("buildah version")
+		var bVersion BuildahVersion
+		_ = json.Unmarshal([]byte(versionJSON), &bVersion)
+		logger.Debug("buildah version: %+v", bVersion)
+		onceBuildahVersion = bVersion
+	})
+	if onceBuildahVersion.GitCommit == "bc5080cc" {
+		return true, nil
+	}
+	return false, errors.New("System's buildah is not match sealos buildah requirements, please uninstall system's buildah, and retry")
```
fanux, line 65: But how to fix this issue, user don't know the gitcommit ID，how to reinstall.

fanux:
1. Users don't know which commitID to download 2. Why other versions of buildah don't work, because our integrated SDK is not new enough, is there a way to be compatible with other versions of buildah 3. The fundamental solution is to not rely on the buildah binary at all

Issue #1982, opened 2022-10-29
Feature:  sealos-system namespace auto generator


cuisongliu:
namespace name is auto generator , not fixed is "sealos-system"

<img width="880" alt="image" src="https://user-images.githubusercontent.com/4150905/198835831-17043fcc-69a5-479e-ae5d-ed4e01dd08ba.png">

delete hard code

PR #1985, opened 2022-10-29
feat: use env instead of hard code

xiao-jay:
Signed-off-by: 晓杰 <2561589453@qq.com>
#1982

### DIFF
--- a/controllers/metering/controllers/metering_controller.go
+++ b/controllers/metering/controllers/metering_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


controllers/metering/controllers/metering_controller.go
```diff
@@ -42,10 +43,19 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+func init() {
+	SealosSystemNamespace = os.Getenv("SEALOS_SYSTEM_NAMESPACE")
```
cuisongliu, line 47: The logic here may be a little unnecessary. This is unlikely to happen from the perspective of deployment.

xiao-jay: agree

cuisongliu (approved):

PR #1995, opened 2022-10-31
feature(main): payment controller disable from env

cuisongliu:


controllers/user/main.go
```diff
@@ -114,12 +116,15 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "Account")
 		os.Exit(1)
 	}
-	if err = (&controllers.PaymentReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "Payment")
-		os.Exit(1)
+	if os.Getenv(pay.CallbackURL) != "" {
```
fanux, line 119: Use payment callbackURL to disable payment controller?
Seems we don't use the callback URL now.

cuisongliu: > Use payment callbackURL to disable payment controller?
> Seems we don't use the callback URL now.
https://github.com/labring/sealos/blob/4a8612e5e8bcba77a8797ea9e3b3c659f36b7074/controllers/user/controllers/payment_controller.go#L76

If there is no callbackURL, the controller startup will fail.

maslow (approved):

Issue #2019, opened 2022-11-02
BUG: terminal hardcoded problem.


zzjin:
### Sealos Version

master

### How to reproduce the bug?

https://github.com/labring/sealos/pull/2013/files#diff-7a592b66e409cc2698ce2bdf92d6772e63d5c70fcfbfe023286361a33090f502R59
Changed From hardcode to env value, so terminal needs to update: 
https://github.com/labring/sealos/blob/main/controllers/terminal/controllers/terminal_controller.go#L200
to match vars from env like using USER_NAMESPACE at kustomization.yaml?

### Operating environment

```markdown
- Sealos version:
- Docker version:
- Kubernetes version:
- Operating system:
- Runtime environment:
- Cluster size:
- Additional information:
```

PR #2028, opened 2022-11-03
fix terminal hardcoded problem

gitccl:
See #2019


cuisongliu:
It may be necessary to support the way users modify this environment variable. You can refer to how the user-controller passes the environment variable, and the cluster image needs to be adjusted.

cuisongliu (approved):

PR #2035, opened 2022-11-03
feature(main): add pyament deploy

cuisongliu:


controllers/user/deploy/README.md
```diff
@@ -8,10 +8,16 @@ sealos build -t docker.io/labring/sealos-user-controller:dev -f Dockerfile .
...
+|`WechatPrivateKey`|Wechat private key|``|
+|`MchID`|Wechat mch id|``|
+|`MchCertificateSerialNumber`|Wechat mch certificate serial number|``|
+|`MchAPIv3Key`|Wechat mch api v3 key|``|
+|`AppID`|Wechat app id|``|
 
 ### How to run
 
 ```shell
+sealos run docker.io/labring/kustomize:v4.5.6
```
fanux, line 21: seems we don't need the callback URL anymore.

cuisongliu: I changed it to `NotifyCallbackURL`. It is actually a callback interface for payment to inform us of the status. I think it's more appropriate to keep it.

fanux: We don't use Notify too, controller will check the payment status every 0.3s, so needn't any notify

fanux: > We don't use Notify too, controller will check the payment status every 0.3s, so needn't notify

cuisongliu: > We don't use Notify too, controller will check the payment status every 0.3s, so needn't any notify

OK，Let me fix this problem.

fanux (approved):

PR #2067, opened 2022-11-08
fix: replace text/template from html/template

fengxsong:

use `text/template` to avoid string escaping.


cuisongliu:
Semgrep / Scan  ，How to solve it ？  @zzjin

cuisongliu (approved):

PR #2128, opened 2022-11-21
fix infra controller status

fanux:


fanux:
Must using `make install` to update the CRD schema. If not, the controller can't update the metadata.

pkg/utils/yaml/yaml.go
```diff
@@ -124,3 +124,8 @@ func Matcher(path string) bool {
 	ext := strings.ToLower(filepath.Ext(path))
 	return ext == ".yaml" || ext == ".yml"
 }
+
+func ShowStructYaml(s interface{}) {
```
zzjin, line 128: Public function not been used.

zzjin (approved):

Issue #2134, opened 2022-11-22
BUG: Failed to pull DigestImage with SHA256


muicoder:
### Sealos Version

v4.1.3

### How to reproduce the bug?

`redis:6.2-alpine`
docker.io/library/redis@sha256:aeaf1a7251c9cdaead1bef13d68aac16b0e22326ecfefdb6079008ffab81065d

### Operating environment

```markdown
- Sealos version: v4.1.3
- Docker version:
- Kubernetes version: v1.24.8
- Operating system: Ubuntu
- Runtime environment: physical machine (4G memory, 2 core cpu, 100G storage)
- Cluster size: 1
- Additional information:
```

### Additional information

![16691063261511](https://user-images.githubusercontent.com/13899572/203269737-3dcbd8c6-9b42-4092-9b00-e1118d91a679.png)

PR #2135, opened 2022-11-22
bug: fixed parse image name and tag

muicoder:
#2134


fanux:
```
cri_server.go:154:3: SA4006: this value of `tag` is never used (staticcheck)
  		tag = imageSplitted[1]
```

fanux (approved):

PR #2145, opened 2022-11-24
feat. Registry auth

lingdie:


service/hub/api/authn.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Authenticator Authentication plugin interface.
+type Authenticator interface {
+	// Authenticate Given a username and a password (plain text), responds with the result or an error.
+	// Error should only be reported if request could not be serviced, not if it should be denied.
+	// A special NoMatch error is returned if the authorizer could not reach a decision,
+	// e.g. none of the rules matched.
+	// Another special WrongPass error is returned if the authorizer failed to authenticate.
+	// Implementations must be goroutine-safe.
+	Authenticate(user string, password PasswordString) (bool, Labels, error)
```
fanux, line 15: What the labels for ?

service/hub/api/authz.go
```diff
@@ -0,0 +1,45 @@
...
+	"net"
+	"strings"
+)
+
+// Authorizer interface performs authorization of the request.
+// It is invoked after authentication, so it can be assumed that the requestor has
+// presented satisfactory credentials for Account.
+// Principally, it answers the question: is this Account allowed to perform these Actions
+// on this Type.Name subject in the give Service?
+type Authorizer interface {
```
fanux, line 14: `authn` and `authz` means what? `Authenticator` interface and `Authorizer` interface what is the relationship？

service/hub/auth/authn.go
```diff
@@ -0,0 +1,45 @@
...
+	api.Authenticator
+}
+
+func (a SealosAuthn) Authenticate(user string, password api.PasswordString) (bool, api.Labels, error) {
+	log.Println("Authenticate for user:", user)
+	client, err := kubernetes.NewKubernetesClientByConfigString(string(password))
+	if err != nil {
+		return false, nil, err
+	}
+	nodes, err := client.Kubernetes().CoreV1().Nodes().List(context.Background(), v1.ListOptions{})
```
fanux, line 26: sealos cloud user kubeconfig don't have the permission to list nodes

service/hub/auth/authn.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		log.Println("Authenticate false")
+		return false, api.Labels{}, nil
+	}
+	log.Println("get nodes: ", nodes)
```
fanux, line 31: You don't need to show the node list in log.

service/hub/auth/authz.go
```diff
@@ -0,0 +1,44 @@
...
+	log.Println("authz plugin init function called")
+}
+
+type SealosAuthz struct {
+	api.Authorizer
+}
+
+func (a SealosAuthz) Authorize(ai *api.AuthRequestInfo) ([]string, error) {
+	log.Println("Authorize for  req:", ai.Name)
+	client, err := kubernetes.NewKubernetesClientByConfigString(ai.Kubeconfig)
```
fanux, line 22: duplicate code！！

service/hub/auth/authz.go
```diff
@@ -0,0 +1,44 @@
...
+
+func init() {
+	log.Println("authz plugin init function called")
+}
+
+type SealosAuthz struct {
+	api.Authorizer
+}
+
+func (a SealosAuthz) Authorize(ai *api.AuthRequestInfo) ([]string, error) {
```
fanux, line 20: duplicate code！

service/hub/server/server.go
```diff
@@ -0,0 +1,373 @@
...
+			ra = hp[1]
+		} else if hp[2] != "" {
+			ra = hp[2]
+		}
+	}
+	res := net.ParseIP(ra)
+	return res
+}
+
+func parseScope(scope string) (string, string, error) {
```
fanux, line 82: scope is what? Provide an example?

service/hub/api/authn.go
```diff
@@ -0,0 +1,36 @@
...
+var NoMatch = errors.New("did not match any rule")
+var WrongPass = errors.New("wrong password for user")
+
+type PasswordString string
+
+func (ps PasswordString) String() string {
+	if len(ps) == 0 {
+		return ""
+	}
+	return "***"
```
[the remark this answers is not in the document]
lingdie: Implement String func overloading of passwords to prevent leakage via print/log

service/hub/server/server.go
```diff
@@ -0,0 +1,373 @@
...
+		realIPPos := as.config.Server.RealIPPos
+		if realIPPos < 0 {
+			realIPPos = len(ips) + realIPPos
+			if realIPPos < 0 {
+				realIPPos = 0
+			}
+		}
+
+		ar.RemoteAddr = strings.TrimSpace(ips[realIPPos])
+		glog.V(3).Infof("Conn ip %s, %s: %s, addr: %s", ar.RemoteAddr, as.config.Server.RealIPHeader, hv, ar.RemoteAddr)
```
fanux, line 113: The logs are not uniform, using both log and glog?

service/hub/server/server.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+	ar.RemoteIP = parseRemoteAddr(ar.RemoteAddr)
+	if ar.RemoteIP == nil {
+		return nil, fmt.Errorf("unable to parse remote addr %s", ar.RemoteAddr)
+	}
+	user, password, haveBasicAuth := req.BasicAuth()
+	if haveBasicAuth {
+		ar.User = user
+		ar.Password = api.PasswordString(password)
+	} else if req.Method == "POST" {
```
fanux, line 126: Maybe you need a webframework. like: https://github.com/emicklei/go-restful

service/hub/api/authn.go
lingdie: labels can carry variables from authn to authz, so it can be more convenient to identify user/org.

service/hub/api/authz.go
lingdie: authn will check user and its password is right or not, authz will range req's scope and check if the user have the right to pull/push image.

service/hub/server/server.go
```diff
@@ -0,0 +1,373 @@
...
+		as.doIndex(rw, req)
+	case req.URL.Path == pathPrefix+"/auth":
+		as.doAuth(rw, req)
+	default:
+		http.Error(rw, "Not found", http.StatusNotFound)
+		return
+	}
+}
+
+func (as *AuthServer) doIndex(rw http.ResponseWriter, req *http.Request) {
```
fanux, line 310: A page to show token?

service/hub/auth/authn.go
lingdie: Only for coding test. Later I will implement this function in its entirety.

service/hub/auth/authn.go
lingdie: Id.

service/hub/server/server.go
lingdie: https://docs.docker.com/registry/spec/auth/scope/#resource-scope-grammar
such as:
https://auth.docker.io/token?service=registry.docker.io&scope=repository:samalba/my-app:pull,push

service/hub/server/server.go
lingdie: ok, I will fix it and use glog.

service/hub/server/server.go
lingdie: return auth server's issuer, it should be add at docker v2 registry config.auth.token.issuer

service/hub/auth/authz.go
lingdie: SealosAuthz is only one kind of authz. I think we should keep it like this.

service/hub/config/server.crt
```diff
@@ -0,0 +1,22 @@
+-----BEGIN CERTIFICATE-----
```
zzjin, line 1: Maybe we must do not commit keys, for develop tests, just create key pair and delete it when tests done, and for prod env, using configmap+secrets  or reading envs from command line will works fine.

lingdie: ok

zzjin:
Add How to deploy `service/hub` into k8s cluster, like deploy or/and config yaml+clsuterfile?

lingdie:
> Add How to deploy `service/hub` into k8s cluster, like deploy or/and config yaml+clsuterfile?

should I consider arm64?

zzjin:
> > Add How to deploy `service/hub` into k8s cluster, like deploy or/and config yaml+clsuterfile?
> 
> should I consider arm64? Any example or limit?

seems your code has no limit and can simply support arm64+amd64 multi-arch images.

Examples:
https://github.com/labring/sealos/blob/main/.github/workflows/services.yml
https://github.com/labring/sealos/blob/main/controllers/infra/Makefile#L68..L87

PR #2148, opened 2022-11-27
Update README.md, use the curl command instead of wget

tanshilingithub:
Update README.md, use the curl command instead of wget, it make install step more clear


tanshilingithub:
curl not work well after my test

PR #2163, opened 2022-11-30
feat: run sealos without root privileges

fengxsong:
This PR contains the following changes:

- refactor/ (migrate) most of the subcommands relies on `buildah`, like `build`/`images`/`rm` , so we can use the most builtin flags, same experience as buildah...
- refactor `imageService`/`clusterService`/`registryService` into one `buildah.Interface`
- remove `pkg/image` module 
- support running sealos without root privileges, to support this feature, `fuse-overlayfs` lib is required


cmd/sealos/cmd/root.go
```diff
@@ -28,12 +28,13 @@ import (
...
 	debug          bool
 	clusterRootDir string
+	runtimeRootDir string
 )
 
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
 	Use:   "sealos",
-	Short: "simplest way install kubernetes tools.",
+	Short: "simplest tool for installing kubernetes tools.",
```
fanux, line 37: - Manage clusters lifecycle
   - Quickly install HA Kubernetes clusters
   - Add / remove nodes
   - Clean the cluster, backup and auto recovering, etc.
- Download and use OCI-compatible distributed applications
   - OpenEBS, MinIO, Ingress, PostgreSQL, MySQL, Redis, etc.
   - Customize your own distributed applications
   - Using Sealosfile to build distributed applications images, saving all dependencies.
   - Push distributed applications images to Docker Hub.
   - Combine multiple applications to build your own cloud platform.
- sealos cloud
   - Have a full public cloud capability, and run it anywhere

pkg/apply/processor/create.go
```diff
@@ -111,29 +108,29 @@ func (c *CreateProcessor) CheckImageType(cluster *v2.Cluster) error {
...
+	err := c.Buildah.Pull(buildah.DefaultPlatform(), "missing", cluster.Spec.Image...)
 	if err != nil {
 		return err
 	}
 	if err = c.CheckImageType(cluster); err != nil {
 		return err
 	}
 	for _, img := range cluster.Spec.Image {
-		clusterManifest, err := c.ClusterManager.Create(fmt.Sprintf("%s-%s", cluster.Name, rand.Generator(8)), img)
+		bderInfo, err := c.Buildah.Create(rand.Generator(8), img)
```
cuisongliu, line 119: fmt.Sprintf("%s-%s", cluster.Name, rand.Generator(8)) , It's better to distinguish the cluster name.

pkg/apply/processor/install.go
```diff
@@ -119,15 +114,15 @@ func (c *InstallProcessor) SyncStatusAndCheck(cluster *v2.Cluster) error {
 }
 
 func (c *InstallProcessor) PreProcess(cluster *v2.Cluster) error {
-	if err := c.RegistryManager.Pull(types.DefaultPlatform(), types.PullPolicyMissing, c.NewImages...); err != nil {
-		return err
-	}
-	ociList, err := c.ImageManager.Inspect(c.NewImages...)
-	if err != nil {
+	if err := c.Buildah.Pull(buildah.DefaultPlatform(), "missing", c.NewImages...); err != nil {
```
cuisongliu, line 117: Miss, is it better to use constants?

cuisongliu:
https://github.com/labring/sealos/issues/1756  This problem needs to be tested, which depends on the binary buildah.

pkg/apply/processor/create.go
fengxsong: > fmt.Sprintf("%s-%s", cluster.Name, rand.Generator(8)) , It's better to distinguish the cluster name.

I've move it to `buildah.Interface`, so `ListContainers` function can only filter working containers that belongs to this cluster.

cmd/sealos/cmd/root.go
fengxsong: `a unified and simplest tool to manage cloudnative toolset` 😄

fanux (approved):

pkg/apply/processor/install.go
```diff
@@ -167,6 +162,17 @@ func (c *InstallProcessor) PreProcess(cluster *v2.Cluster) error {
 	return nil
 }
 
+func replaceOrInsert(ss []string, s string) []string {
+	var ret []string
+	for i := range ss {
+		if ss[i] != s {
```
cuisongliu, line 168: This modification may change the order. Wouldn't it better to replace the content in ss directly?

if found: 
ss[i]==s 
   ss[i]=s
if not found:
   append(ss,s)

cuisongliu (approved):

PR #2164, opened 2022-11-30
fixed set system disk to 40G by default, the device name must be /dev…

fanux:
…/xvda


fanux:
Fixed not enough disk bug

cuisongliu (approved):

PR #2176, opened 2022-12-02
feat. sealos login hub.sealos.io by kubeconfig.

lingdie:
sealos login hub.sealos.uo by kubeconfig.

### DIFF
--- a/pkg/buildah/login.go
+++ b/pkg/buildah/login.go
@@ -17,6 +17,7 @@ package buildah
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 
 	"github.com/containers/buildah/pkg/parse"
@@ -27,9 +28,6 @@ import (
 
 	fileutil "github.com/labring/sealos/pkg/utils/file"
 )
-
-const DefaultKubeConfigDir = ".sealos/sealos.io"
-const DefaultRegistry = "hub.sealos.io"
 
 type loginReply struct {
 	loginOpts  auth.LoginOptions
@@ -70,44 +68,39 @@ func newLoginCommand() *cobra.Command {
 		Long:  loginDescription,
 
 		PreRunE: func(cmd *cobra.Command, args []string) (err error) {
-			if len(args) == 0 {
-				args = append(args, DefaultRegistry)
-			}
-
-			if args[0] == DefaultRegistry {
-				if opts.loginOpts.Username != "" {
-					return fmt.Errorf("the hub.sealos.io registry can only be logged in by kubeconfig")
-				}
-			}
-
 			// login by kubeconfig
 			if opts.kubeconfig != "" {
-				sealoskubeconfdir := fmt.Sprintf("%s/%s", os.Getenv("HOME"), DefaultKubeConfigDir)
-				err = fileutil.MkDirs(sealoskubeconfdir)
-				if err != nil {
-					return err
-				}
-
-				sealoskubeconfpath := fmt.Sprintf("%s/%s", sealoskubeconfdir, "config")
-
-				// copy kubeconfig to ${HOME}/.sealos/sealos.io/config
-				err = fileutil.Copy(opts.kubeconfig, sealoskubeconfpath)
-				if err != nil {
-					return err
-				}
-
 				// set user/password
 				var passwordb []byte
-				if passwordb, err = fileutil.ReadAll(sealoskubeconfpath); err != nil {
+				if passwordb, err = fileutil.ReadAll(opts.kubeconfig); err != nil {
 					return err
 				}
 				// username is current context kubeconfig user id
-				username, err := GetCurrentUserFromKubeConfig(sealoskubeconfpath)
+				username, err := GetCurrentUserFromKubeConfig(opts.kubeconfig)
 				if err != nil {
 					return err
 				}
 				opts.loginOpts.Username = username
 				opts.loginOpts.Password = string(passwordb)
+
+				// config will be copyed to $(HOME)/.sealos/$(args[0]).HOST/$(user)/config
+				registryUrl, err := url.Parse(args[0])
+				if err != nil {
+					return err
+				}
+				if registryUrl == nil {
+					return errors.New("registry url is not legal")
+				}
+				sealosKubeConfdir := fmt.Sprintf("%s/%s/%s/%s", os.Getenv("HOME"), ".sealos", registryUrl.Host, username)
+				err = fileutil.MkDirs(sealosKubeConfdir)
+				if err != nil {
+					return err
+				}
+				sealosKubeconfPath := fmt.Sprintf("%s/%s", sealosKubeConfdir, "config")
+				err = fileutil.Copy(opts.kubeconfig, sealosKubeconfPath)
+				if err != nil {
+					return err
+				}
 			}
 			return nil
 		},
@@ -128,8 +121,7 @@ func loginCmd(c *cobra.Command, args []string, iopts *loginReply) error {
 		return errors.New("too many arguments, login takes only 1 argument")
 	}
 	if len(args) == 0 {
-		args = append(args, DefaultRegistry)
-		fmt.Println("login to default registry: hub.sealos.io.")
+		return errors.New("please specify a registry to login to")
 	}
 
 	if err := setXDGRuntimeDir(); err != nil {
--- a/pkg/buildah/login.go
+++ b/pkg/buildah/login.go
@@ -69,8 +69,9 @@ func newLoginCommand() *cobra.Command {
 		Long:  loginDescription,
 
 		PreRunE: func(cmd *cobra.Command, args []string) (err error) {
-			// login by kubeconfig
-			if opts.kubeconfig != "" {
+			if opts.kubeconfig == "" {
+				return
+			} else {
 				// set user/password
 				var passwordb []byte
 				if passwordb, err = fileutil.ReadAll(opts.kubeconfig); err != nil {
--- a/pkg/buildah/login.go
+++ b/pkg/buildah/login.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


pkg/buildah/login.go
```diff
@@ -75,7 +128,8 @@ func loginCmd(c *cobra.Command, args []string, iopts *loginReply) error {
 		return errors.New("too many arguments, login takes only 1 argument")
 	}
 	if len(args) == 0 {
-		return errors.New("please specify a registry to login to")
+		args = append(args, DefaultRegistry)
+		fmt.Println("login to default registry: hub.sealos.io.")
```
zzjin, line 132: Seems not necessary, as explained above.

pkg/buildah/login.go
```diff
@@ -59,14 +68,58 @@ func newLoginCommand() *cobra.Command {
 		Use:   "login",
 		Short: "Login to a container registry",
 		Long:  loginDescription,
+
+		PreRunE: func(cmd *cobra.Command, args []string) (err error) {
+			if len(args) == 0 {
+				args = append(args, DefaultRegistry)
+			}
+
+			if args[0] == DefaultRegistry {
```
zzjin, line 77: We can support any host wether they support `sealos` login or not.
Just parse user inputs, and do login, and save response.

If user want to login to `docker.io`, they can just use command: `sealos login docker.io -u xxx -p xxx` as usual, alone with user can login into `hub.sealos.io` witch currently support `sealos login hub.sealos.io -u xxx -k path/to/file`.

With this design, we can both support old behavior and plus support `kubeconfig` login. Further more, once our `hub.sealos.io` updated support  username+password login(which we can do but not do right now), users can login any registries as they need using the same method.

cuisongliu:
I don't think it's appropriate to modify buildah's code directly. Is it necessary to add an applier layer for modification? This is the underlying logic. If you use this module in the startup place, these functions will be turned on by default.

pkg/buildah/login.go
lingdie: > sealos login hub.sealos.io -u xxx -k path/to/file.
here, we can parse kubeconfig and get user.

lingdie:
> I don't think it's appropriate to modify buildah's code directly. Is it necessary to add an applier layer for modification? This is the underlying logic. If you use this module in the startup place, these functions will be turned on by default.

We are not changing the functionality of the original function, we are just adding a flag to it. Originally there was pkg/image that did what you are talking about, but it was removed.

pkg/buildah/login.go
```diff
@@ -48,6 +54,8 @@ func (opts *loginReply) RegisterFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&opts.tlsVerify, "tls-verify", opts.getLogin, "require HTTPS and verify certificates when accessing the registry. TLS verification cannot be used when talking to an insecure registry.")
 	fs.BoolVar(&opts.getLogin, "get-login", opts.tlsVerify, "return the current login user for the registry")
 	fs.AddFlagSet(auth.GetLoginFlags(&opts.loginOpts))
+	// e.g sealos login --kubeconfig /root/.kube/config hub.sealos.io
+	fs.StringVarP(&opts.kubeconfig, "kubeconfig", "k", opts.kubeconfig, "Login to sealos registry: hub.sealos.io by kubeconfig")
```
fanux, line 58: set a default value `~/.kube/config`

pkg/buildah/login.go
```diff
@@ -59,14 +67,50 @@ func newLoginCommand() *cobra.Command {
 		Use:   "login",
 		Short: "Login to a container registry",
 		Long:  loginDescription,
+
+		PreRunE: func(cmd *cobra.Command, args []string) (err error) {
+			// login by kubeconfig
+			if opts.kubeconfig != "" {
```
fanux, line 73: Return fast:
```
if opts.kubeconfig == "" {
    return ...
}
```

And if set the default value, opt.kubeconfig will not be ""

pkg/buildah/login.go
```diff
@@ -59,14 +67,50 @@ func newLoginCommand() *cobra.Command {
...
+				// set user/password
+				var passwordb []byte
+				if passwordb, err = fileutil.ReadAll(opts.kubeconfig); err != nil {
```
fanux, line 76: The all kubeconfig content as password?

lingdie: yes.

pkg/buildah/login.go
```diff
@@ -59,14 +67,51 @@ func newLoginCommand() *cobra.Command {
 		Use:   "login",
 		Short: "Login to a container registry",
 		Long:  loginDescription,
+
+		PreRunE: func(cmd *cobra.Command, args []string) (err error) {
+			if opts.kubeconfig == "" {
+				return
+			} else {
```
zzjin, line 74: This if statement can simply delete `else`, for faster code return and better code Indent .

pkg/buildah/login.go
```diff
@@ -90,3 +135,29 @@ func loginCmd(c *cobra.Command, args []string, iopts *loginReply) error {
...
+		return "", fmt.Errorf("failed to find current context %s", config.CurrentContext)
+	}
+	return expectedCtx.AuthInfo, nil
+}
+
+// If the specified string starts with http{s} it is replaced with it's
+// host[:port] parts; everything else is stripped. Otherwise, the string is
+// returned as is.
+func replaceURLByHostPort(repository string) (string, error) {
+	if !strings.HasPrefix(repository, "https://") && !strings.HasPrefix(repository, "http://") {
```
zzjin, line 155: Not as safe if passed by urls like 'ftp://sealos.io/../../root/.ssh/auth'.
 Maybe call `url.ParseRequestURI` and check return err.

pkg/buildah/login.go
```diff
@@ -59,14 +67,51 @@ func newLoginCommand() *cobra.Command {
...
+				if err != nil {
+					return err
+				}
+				sealosKubeConfdir := fmt.Sprintf("%s/%s/%s/%s", os.Getenv("HOME"), ".sealos", registryHost, username)
+				err = fileutil.MkDirs(sealosKubeConfdir)
+				if err != nil {
+					return err
+				}
+				sealosKubeconfPath := fmt.Sprintf("%s/%s", sealosKubeConfdir, "config")
+				err = fileutil.Copy(opts.kubeconfig, sealosKubeconfPath)
```
zzjin, line 99: Seems we must wait login command succeeded, and then do the real copy action?

pkg/buildah/login.go
zzjin: Seems not necessary. If we downloaded kubeconfig from our cloud site, the file will be in any location.

There is only one statement that`~/.kube/config` exists, witch is users using `sealos run` to start and manage a cluster, witch at this statement, they are not likely do login or push images, but  just pull and run sealos images.

pkg/buildah/login.go
lingdie: ok...

pkg/buildah/login.go
lingdie: I will move it to PostRunE.

pkg/buildah/login.go
lingdie: Any legitimate url could be a registry, and it's hard to use the filesystem to distinguish between different registries or different sealos-cloud-instance.
For example:
- https://hub.sealos.io/registry1
- https://hub.sealos.io/registry2/part2
or without scheme
- hub.sealos.io:5000
or
- yy.hub.sealos.io.cn


I was wondering if we should also maintain a jason file like docker, e.g.
```jason
{
	"registrys": {
		"hub.sealos.io:35000": {
			"kubeconfig": "...",
                        "user": "...",
		},
	},
        "clouds":{
        }
}
```
so, we can just edit and read it to get kubeconfig.


Or Based on userid uniqueness just maintain file like:$(HOME)/.sealos/users/$(userid)/config
The probability of different sealos generating the same userid and using it on the same cluster is very low.

pkg/buildah/login.go
```diff
@@ -59,14 +67,53 @@ func newLoginCommand() *cobra.Command {
 		Use:   "login",
 		Short: "Login to a container registry",
 		Long:  loginDescription,
+
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if opts.kubeconfig == "" {
+				return nil
```
fanux, line 73: need return not found error~

pkg/buildah/login.go
zzjin: Use just `hostname` for path key is enough.

* http://hub.sealos.io/registry1 -> hub.sealos.io
* https://hub.sealos.io/registry2/part2 -> hub.sealos.io
* hub.sealos.io:5000 -> hub.sealos.io:5000
* yy.hub.sealos.io.cn -> yy.hub.sealos.io.cn

`http` and `https` refers to one domain.

We do not support one hotname with two or more registry, we can use:
* `registry1.hub.sealos.io`
* `registry2.hub.sealos.io` 

instead.

pkg/buildah/login.go
```diff
@@ -59,14 +66,53 @@ func newLoginCommand() *cobra.Command {
...
+			registryHost, _, err := parseRawURL(args[0])
+			if err != nil {
+				return err
+			}
+			sealosKubeConfdir := fmt.Sprintf("%s/%s/%s/%s", os.Getenv("HOME"), ".sealos", registryHost, opts.loginOpts.Username)
+			if err := fileutil.MkDirs(sealosKubeConfdir); err != nil {
+				return err
+			}
+			sealosKubeconfPath := fmt.Sprintf("%s/%s", sealosKubeConfdir, "config")
+			if err := fileutil.Copy(opts.kubeconfig, sealosKubeconfPath); err != nil {
```
zzjin, line 105: Named `Copy` will override exists file, add comment to spec it.

zzjin (approved):

PR #2203, opened 2022-12-07
fix: reconcile instance count

xiaohan1202:
Signed-off-by: xiaohan <962479661@qq.com>


controllers/infra/config/samples/infra_v1_infra.yaml
```diff
@@ -29,7 +29,7 @@ spec:
         - capacity: 21
           volumeType: gp3
           type: "data"
-    - roles: [ node ]
+    - roles: [ node1 ]
```
fanux, line 32: node roles must contains `node`

fanux (approved):

PR #2207, opened 2022-12-07
fix: move IsRootless out of pkg/buildah

fengxsong:

In case some other go module require pkg/ssh, so move the only necessary function out of pkg/buildah.


zzjin (approved):

cubxxw:
buildah is built under rootless condition through unshare. Does sealos satisfy `sealos run` under rootless condition?

PR #2229, opened 2022-12-09
feature(main): add delete build file scan

cuisongliu:

### DIFF
--- a/pkg/buildah/build.go
+++ b/pkg/buildah/build.go
@@ -17,6 +17,10 @@ package buildah
 import (
 	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/containers/buildah/define"
 
 	"github.com/containers/buildah/imagebuildah"
 	buildahcli "github.com/containers/buildah/pkg/cli"
@@ -98,6 +102,34 @@ func buildCmd(c *cobra.Command, inputArgs []string, sopts saveOptions, iopts bui
 		defer iopts.Logwriter.Close()
 	}
 
+	if len(iopts.File) == 0 {
+		ctxDir, err := getContextDir(inputArgs)
+		if err != nil {
+			return err
+		}
+		getFile := func(filename string) (string, error) {
+			file := filepath.Join(ctxDir, filename)
+			fileInfo, err := os.Stat(file)
+			if err != nil {
+				return "", fmt.Errorf("cannot find %s in context directory: %w", filename, err)
+			}
+			// The file exists, now verify the correct mode
+			if mode := fileInfo.Mode(); mode.IsRegular() {
+				return file, nil
+			}
+			return "", fmt.Errorf("assumed %s is a file but it's not", filename)
+		}
+
+		defaultFileNames := []string{"Sealfile", "Kubefile", "Dockerfile", "Containerfile"}
+		for _, name := range defaultFileNames {
+			if foundFile, err := getFile(name); err == nil && foundFile != "" {
+				iopts.File = append(iopts.File, foundFile)
+			}
+		}
+		if len(iopts.File) == 0 {
+			return fmt.Errorf("cannot find any of %v in context directory", strings.Join(defaultFileNames, ", "))
+		}
+	}
 	options, containerfiles, removeAll, err := buildahcli.GenBuildOptions(c, inputArgs, iopts)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


pkg/buildah/common.go
```diff
@@ -238,3 +241,36 @@ Flags:
 {{end}}
 `
 }
+
+func getContextDir(inputArgs []string) (string, error) {
```
fengxsong, line 245: Since this function is only used by the `buildCmd` method, it is more appropriate to move it to the `build.go` file?

pkg/buildah/build.go
```diff
@@ -98,6 +99,39 @@ func buildCmd(c *cobra.Command, inputArgs []string, sopts saveOptions, iopts bui
 		defer iopts.Logwriter.Close()
 	}
 
+	if len(iopts.File) == 0 {
```
fengxsong, line 102:
```golang
	if len(iopts.File) == 0 {
		ctxDir, err := getContextDir(inputArgs)
		if err != nil {
			return err
		}
		getFile := func(filename string) (string, error) {
			file := filepath.Join(ctxDir, filename)
			fileInfo, err := os.Stat(file)
			if err != nil {
				return "", fmt.Errorf("cannot find %s in context directory: %w", filename, err)
			}
			// The file exists, now verify the correct mode
			if mode := fileInfo.Mode(); mode.IsRegular() {
				return file, nil
			} 
			return "", fmt.Errorf("assumed %s is a file but it's not", filename)
		}

		defaultFileNames := []string{"Dockerfile", "Containerfile", "Kubefile", "Sealfile"}
		for _, name := range defaultFileNames {
			if foundFile, err := getFile(ctx, name); err == nil && foundFile != "" {
					iopts.File = []string{foundFile}
				}
			}
		}
                 if len(iopts.File) == 0 {
                         return fmt.Errorf("cannot find any of %v in context directory", strings.Join(defaultFileNames, ", "))
                }
	}
```https://docs.github.com/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax

File: pkg/buildah/build.go [hunk 3]
```diff
@@ -98,6 +102,39 @@ func buildCmd(c *cobra.Command, inputArgs []string, sopts saveOptions, iopts bui
...
+			// The file exists, now verify the correct mode
+			if mode := fileInfo.Mode(); mode.IsRegular() {
+				foundCtrFile = ctrfile
+			} else {
+				return "", fmt.Errorf("assumed %s %q is not a file", file, ctrfile)
+			}
+			return foundCtrFile, nil
+		}
+
+		defaultFileNames := []string{"Dockerfile", "Containerfile", "Kubefile", "Sealfile"}
```
fanux, line 126: Only use Sealosfile in the feature.

cuisongliu: > Only use Sealosfile in the feature.

Do we need to be compatible with the previous logic?

pkg/buildah/build.go
```diff
@@ -98,6 +102,39 @@ func buildCmd(c *cobra.Command, inputArgs []string, sopts saveOptions, iopts bui
 		defer iopts.Logwriter.Close()
 	}
 
+	if len(iopts.File) == 0 {
```
fengxsong, line 105: L105-L132
```golang
	if len(iopts.File) == 0 {
		ctxDir, err := getContextDir(inputArgs)
		if err != nil {
			return err
		}
		getFile := func(filename string) (string, error) {
			file := filepath.Join(ctxDir, filename)
			fileInfo, err := os.Stat(file)
			if err != nil {
				return "", fmt.Errorf("cannot find %s in context directory: %w", filename, err)
			}
			// The file exists, now verify the correct mode
			if mode := fileInfo.Mode(); mode.IsRegular() {
				return file, nil
			}
			return "", fmt.Errorf("assumed %s is a file but it's not", filename)
		}

		defaultFileNames := []string{"Dockerfile", "Containerfile", "Kubefile", "Sealfile"}
		for _, name := range defaultFileNames {
			if foundFile, err := getFile(name); err == nil && foundFile != "" {
				iopts.File = append(iopts.File, foundFile)
			}
		}
		if len(iopts.File) == 0 {
			return fmt.Errorf("cannot find any of %v in context directory", strings.Join(defaultFileNames, ", "))
		}
	}
```

fengxsong (approved):
LGTM

PR #2236, opened 2022-12-10
fix: get the full root command name

fengxsong:

In case buildah's subcommands are registered to a sub command, so get the full root command name.

### DIFF
--- a/pkg/buildah/build.go
+++ b/pkg/buildah/build.go
@@ -32,12 +32,12 @@ import (
 )
 
 func newBuildCommand() *cobra.Command {
-	buildDescription := `
-  Builds an OCI image using instructions in one or more Containerfiles.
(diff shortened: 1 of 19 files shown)


pkg/buildah/build.go
```diff
@@ -32,12 +32,12 @@ import (
 )
 
 func newBuildCommand() *cobra.Command {
-	buildDescription := `
-  Builds an OCI image using instructions in one or more Containerfiles.
+	buildDescription := fmt.Sprintf(`
+  Builds an OCI image using instructions in one or more Containerfiles/Dockerfiles/Kubefiles/Sealfiles.
```
cuisongliu, line 36: Adjust the order?

Sealfile, Kubefile,Dockerfile,Containerfile

cuisongliu (approved):

PR #2244, opened 2022-12-12
use admisswebhook sdk to mutate and validate image create, update, delete.

lingdie:
use admisswebhook sdk to mutate and validate image create, update, delete.


controllers/imagehub/api/v1/image_webhook.go
```diff
@@ -17,69 +17,127 @@ limitations under the License.
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
 // log is for logging in this package.
 var imagelog = logf.Log.WithName("image-resource")
 
+const saPrefix = "system:serviceaccount:user-system:"
```
lingdie, line 35: I need this to replace sa.name.prefix to "" and compare it with org.spec.manager.

controllers/imagehub/api/v1/image_webhook.go
```diff
@@ -17,69 +17,127 @@ limitations under the License.
...
+	// replace it and compare
+	userName := strings.Replace(req.UserInfo.Username, saPrefix, "", -1)
+	for _, usr := range org.Spec.Manager {
+		if usr == userName {
+			return nil
+		}
+	}
+	imagelog.Info("denied", "image name", i.Name)
+	return fmt.Errorf("denied, you are not one of organization %s managers", i.Spec.Name.GetOrg())
 }
```
lingdie, line 143: this is the func to get and check org.spec.manager.

controllers/imagehub/api/v1/image_webhook.go
zzjin: Since user controller may be deploy to any system namespace, so you need up update this to read ns name from env like: `--env USER_NAMESPACE=xxx`

zzjin (approved):

PR #2255, opened 2022-12-13
add imagecrdbuilder In push logic

cdjianghan:
add imageCrdBuilder
    split some function to make the logic clear
    fit new buildah implement
    add image id arch and size read


cdjianghan:
* the Run function only print warning rather than return err when something wrong happen
* add read app content pipeline if there will have more content read in the future
* imagehub check now just use  flags temporary

fanux (approved):

pkg/buildah/imageCrdBuilder.go
```diff
@@ -0,0 +1,304 @@
...
+	var size int64
+	size = manifest.Config.Size * 1000
+	for _, layer := range manifest.Layers {
+		size += layer.Size
+	}
+	icb.AppConfig.Spec.DetailInfo.Size = size
+}
+
+// ReadActions read all action in /app dir
+func (icb *ImageCRDBuilder) ReadActions(MountPoint string) error {
```
zzjin, line 259: Split-out actions for this pr?

zzjin (approved):

PR #2257, opened 2022-12-13
feat. add repo webhook to set default labels and validate it.

lingdie:
feat. add repo webhook to set default labels and validate it.


controllers/imagehub/api/v1/repository_webhook.go
```diff
@@ -17,70 +17,131 @@ limitations under the License.
...
+	repositorylog.Info("org info", "org", org)
+	repositorylog.Info("getting req from ctx")
+	req, err := admission.RequestFromContext(ctx)
+	if err != nil {
+		repositorylog.Info("get request from context error when validate", "repository name", r.Name)
+		return err
+	}
+	repositorylog.Info("checking user", "user", req.UserInfo.Username)
+	// get sa namespace prefix, prefix format is like: "system:serviceaccount:user-system:"
+	namespacePrefix := fmt.Sprintf("%s:%s:", saPrefix, getUserNamespace())
```
zzjin, line 132: This `saPrefix` value is shared over multiple webhooks, it should move to standalone const file, like move into `common_types.go`

lingdie: ok

zzjin (approved):

PR #2264, opened 2022-12-14
Update SEALOS_SYS_KUBE_VERSION usage.

zzjin:


cuisongliu (approved):

pkg/template/funcmap.go
```diff
@@ -181,3 +184,22 @@ func ipAt(s string, idx uint32) string {
...
+func semverCompare(constraint, version string) (bool, error) {
+	c, err := sv2.NewConstraint(constraint)
+	if err != nil {
+		return false, err
+	}
+
+	v, err := sv2.NewVersion(version)
+	if err != nil {
+		// nosemgrep: dgryski.semgrep-go.nilerr.return-nil
+		return false, nil
```
cuisongliu, line 201: add //nolint:nilerr

cuisongliu (changes requested):

pkg/template/funcmap.go
```diff
@@ -50,6 +51,8 @@ func funcMap() template.FuncMap {
 		"fromJsonArray": fromJSONArray,
 		"ipNet":         ipNet,
 		"ipAt":          ipAt,
+
+		"semverCompare": semverCompare, // custom semver compare to avoid error
```
fengxsong, line 55: don't override the default template function, cause this method may be called in other packages， such as `pkg/buildimage`

zzjin: But this will cause error, when `semver` is not valid. What I override is return (false,nil) to continue if passed in version is invalid statement, but not (false, err) to break all progress.
Any other suggesttions?

cuisongliu (approved):

PR #2265, opened 2022-12-15
feature(main): env compatible

cuisongliu:


pkg/env/env.go
```diff
@@ -16,13 +16,15 @@ package env
...
 // nosemgrep: go.lang.security.audit.xss.import-text-template.import-text-template
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
-	"text/template"
 
+	template2 "github.com/labring/sealos/pkg/template"
```
zzjin, line 25: Here is nomore use of default text/template, so we can just import as template?

zzjin (approved):

Issue #1973, opened 2022-10-28
BUG: sealos build not Parsing some helm image


willzhang:
**Which command or component**

sealos build

**The Description of the bug**

Describe the bug.

**How to reproduce（pictures can be attached if necessary）**

```
# sealos  build -t registry.cn-shenzhen.aliyuncs.com/cnmirror/apisix:2.15.0 .2022-10-28T13:31:27 info lookup in path charts
2022-10-28T13:31:27 info sub chart is apisix
2022-10-28T13:31:29 warn if you access private registry,you must be 'sealos login' or 'buildah login'
2022-10-28T13:31:29 info pull images [docker.io/apache/apisix:2.15.0-alpine docker.io/bitnami/etcd:3.5.4-debian-11-r14 docker.io/library/busybox:1.28] for platform is linux/amd64
Pulling image: docker.io/apache/apisix:2.15.0-alpine
Pulling image: docker.io/bitnami/etcd:3.5.4-debian-11-r14
Pulling image: docker.io/library/busybox:1.28
```

**Test Result**
just Parsing  3 images
```
Pulling image: docker.io/apache/apisix:2.15.0-alpine
Pulling image: docker.io/bitnami/etcd:3.5.4-debian-11-r14
Pulling image: docker.io/library/busybox:1.28
```

**What you expected to happen**

it should parsing dashboard and ingress-controller too, because of `ingress-controller.enabled=true,dashboard.enabled=true`

```
helm upgrade -i apisix charts/apisix -n ingress-apisix --create-namespace --set gateway.type=NodePort,ingress-controller.enabled=true,dashboard.enabled=true,ingress-controller.config.apisix.serviceNamespace=ingress-apisix
```

parsing all need images
```
# helm template apisix apisix/apisix -n ingress-apisix --create-namespace --set gateway.type=NodePort,ingress-controller.enabled=true,dashboard.enabled=true,ingress-controller.config.apisix.serviceNamespace=ingress-apisix | grep image:
          image: "apache/apisix-dashboard:2.13-alpine"
          image: busybox:1.28
          image: "apache/apisix-ingress-controller:1.5.0"
          image: "apache/apisix:2.15.0-alpine"
          image: busybox:1.28
          image: docker.io/bitnami/etcd:3.5.4-debian-11-r14
      image: busybox
```

**Operating environment**

- Docker version:   eg. 19.03
- Kubernetes version: eg.kube1.16.tar.gz
- Sealos version: v4.1.4-rc1
[...]

PR #2272, opened 2022-12-15
fix: missing images from chart dependencies

wxharry:
Signed-off-by: Harry <xiaohanwu12@gmail.com>

#1973

### DIFF
--- a/pkg/buildimage/chart.go
+++ b/pkg/buildimage/chart.go
@@ -43,11 +43,27 @@ func ParseChartImages(chartPath string) ([]string, error) {
 	if len(subChartPaths) == 0 {
 		return []string{}, nil // if chartPath not exist, return []
 	}
+
 	allImages := sets.NewString()
-	for _, subChartPath := range subChartPaths {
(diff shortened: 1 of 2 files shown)


fengxsong:
No need to manually do this. You can check out codes at L101-L112

pkg/buildimage/chart.go
```diff
@@ -43,12 +43,28 @@ func ParseChartImages(chartPath string) ([]string, error) {
...
 		return []string{}, nil // if chartPath not exist, return []
 	}
+
 	allImages := sets.NewString()
-	for _, subChartPath := range subChartPaths {
+	logger.Info(subChartPaths)
+	// add dependencies to subChartPaths list while iterating the list
+	for i := 0; i < len(subChartPaths); i++ {
+		subChartPath := subChartPaths[i]
+		logger.Info("loop iteration counter", i)
```
cuisongliu, line 52: This log should be debug level

fengxsong:
As we discussed, no changes need to be made.

PR #2275, opened 2022-12-16
add registry deply config.

lingdie:


zzjin:
deploy readme needs real content，and also with common f&q.

zzjin (changes requested):
add readme content.

zzjin (approved):

PR #2276, opened 2022-12-16
fix: only set default --platform flag when it's not changed

fengxsong:

### DIFF
--- a/pkg/buildah/create.go
+++ b/pkg/buildah/create.go
@@ -43,12 +43,15 @@ func newCreateCmd() *cobra.Command {
 			if !IsRootless() {
 				return nil
 			}
-			shell, shellSet := os.LookupEnv("SHELL")
-			if !shellSet {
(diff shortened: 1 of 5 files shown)


pkg/buildah/create.go
```diff
@@ -43,12 +43,15 @@ func newCreateCmd() *cobra.Command {
 			if !IsRootless() {
 				return nil
 			}
-			shell, shellSet := os.LookupEnv("SHELL")
-			if !shellSet {
-				logger.Error("no command specified and no `SHELL` env set")
-				os.Exit(1)
+			args = args[1:]
+			if len(args) < 1 {
+				shell, shellSet := os.LookupEnv("SHELL")
```
fanux, line 48: add some docs for this ENV?

fengxsong: `SHELL` is the user's shell. in most case it's automatically set by [login](http://man.openbsd.org/login.1#SHELL).

PR #2283, opened 2022-12-20
feature: add merge feature

cuisongliu:


pkg/buildah/merge.go
```diff
@@ -0,0 +1,155 @@
...
+				Architecture: arch,
+				OS:           oss,
+				Variant:      variant,
+			}
+			var err error
+			buildahInfo, err = mergeImagesWithScratchContainer(tag[0], args, platform)
+			if err != nil {
+				return err
+			}
+			if err := setDefaultFlagsWithSetters(cmd, setDefaultSaveImageFalseFlag); err != nil {
```
fengxsong, line 76: no necessary to define a function since is not a very common use. just `return setDefaultFlagsWithSetters(cmd, setDefaultFlagIfNotChanged(c, "save-image", "false"))`.

pkg/buildah/common.go
```diff
@@ -68,10 +69,46 @@ func setDefaultTLSVerifyFlag(c *cobra.Command) error {
...
+
+func getVariantFromFlags(c *cobra.Command) string {
+	variant, err := c.Flags().GetString("variant")
+	if err != nil {
+		return ""
+	}
+	return variant
+}
+
+func getTagFromFlags(c *cobra.Command) []string {
```
fengxsong, line 104: getTagsFromFlags

pkg/buildah/merge.go
cuisongliu: done

pkg/buildah/common.go
cuisongliu: done

fengxsong (approved):

PR #2287, opened 2022-12-21
fix: more friendly error output when running rootless mode 

fengxsong:
1. more friendly error output when running rootless mode without required packages like `uidmap`/`fuse-overlayfs`
2. use builtin `cobra.Command.CommandPath()` to replace `getFullCmdName`

### DIFF
--- a/cmd/sealos/cmd/root.go
+++ b/cmd/sealos/cmd/root.go
@@ -59,7 +59,7 @@ func init() {
 	fs.StringVar(&runtimeRootDir, "sealos-root", constants.DefaultRuntimeRootDir, "root directory for sealos actions")
 	_ = fs.MarkHidden("sealos-root")
 	// add unrelated command names that don't required buildah sdk.
-	buildah.AddUnrelatedCommandNames("docs", "exec", "gen", "scp", "version")
+	buildah.AddUnrelatedCommandNames("cert", "docs", "exec", "gen", "scp", "version")
 }
 
 func onBootOnDie() {
--- a/cmd/sealos/cmd/root.go
+++ b/cmd/sealos/cmd/root.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/labring/sealos/pkg/buildah"
 	"github.com/labring/sealos/pkg/constants"
 	"github.com/labring/sealos/pkg/utils/file"
 	"github.com/labring/sealos/pkg/utils/logger"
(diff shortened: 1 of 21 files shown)


cmd/sealos/cmd/root.go
```diff
@@ -57,6 +58,8 @@ func init() {
 	_ = fs.MarkHidden("cluster-root")
 	fs.StringVar(&runtimeRootDir, "sealos-root", constants.DefaultRuntimeRootDir, "root directory for sealos actions")
 	_ = fs.MarkHidden("sealos-root")
+	// add unrelated command names that don't required buildah sdk.
+	buildah.AddUnrelatedCommandNames("docs", "exec", "gen", "scp", "version")
```
cuisongliu, line 62: add cert

cmd/sealos/cmd/root.go
```diff
@@ -57,6 +58,8 @@ func init() {
 	_ = fs.MarkHidden("cluster-root")
 	fs.StringVar(&runtimeRootDir, "sealos-root", constants.DefaultRuntimeRootDir, "root directory for sealos actions")
 	_ = fs.MarkHidden("sealos-root")
+	// add unrelated command names that don't required buildah sdk.
+	buildah.AddUnrelatedCommandNames("cert", "docs", "exec", "gen", "scp", "version")
```
zzjin, line 62: Move this params to const ahead of this file?

pkg/buildah/buildah.go
```diff
@@ -107,7 +108,7 @@ func RegisterGlobalFlags(fs *pflag.FlagSet) error {
 var (
 	globalFlagResults globalFlags
 	rootCmd           *cobra.Command
-	rootCmdName       string
+	unrelatedCommands = sets.NewString("version")
```
zzjin, line 111: `k8s.io/apimachinery/pkg/util/sets`'s `String` is deprecated. Use their new: `https://github.com/kubernetes/apimachinery/blob/master/pkg/util/sets/set.go`

zzjin: Maybe just new empty string is enough.(Suppress const vars.)

fanux (approved):

zzjin (changes requested):
Need some update.

pkg/buildah/buildah.go
fengxsong: OK

cmd/sealos/cmd/root.go
fengxsong: Each subcommands can call function `buildah.AddUnrelatedCommandNames` separately, so I don't see the need to make it as constant. @cuisongliu what do you suggest?

zzjin (approved):

PR #2291, opened 2022-12-21
feat.add image keywords as repo label.

lingdie:
feat.add image keywords as repo label.


controllers/imagehub/api/v1/repository_types.go
```diff
@@ -103,6 +104,16 @@ func (r *Repository) getName() string {
 	return r.Name
 }
 
+func (r *Repository) genKeywordsLabels(img *Image) {
+	if r.Labels == nil {
+		r.Labels = make(map[string]string)
+	}
+	for _, keyword := range img.Spec.DetailInfo.Keywords {
```
zzjin, line 111: the name is only supported by: 
```
The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between. 
```
Direct use of `DetailInfo.Keywords` will fail duo to some wrong words?

lingdie: solved. use a map to check.

controllers/imagehub/api/v1/common_types.go
```diff
@@ -28,6 +31,19 @@ const (
 	defaultImagehubNamespace = "imagehub-system"
 )
 
+func GetSupportedKeywords() []string {
+	return []string{"Kubernetes", "Storage", "Network", "Database", "Monitoring", "Logging", "Dashboard", "MQ", "Platform", "GPU", "GitOps"}
+}
+
+var supportedKeywordsMap map[string]struct{}
```
zzjin, line 38:
```go
var supportedKeywordsMap = func() map[string]struct{} {
return xxxx;
}()
```
Make map var simper at build time not init runtime.

lingdie: done.

controllers/imagehub/api/v1/repository_types.go
```diff
@@ -103,6 +104,19 @@ func (r *Repository) getName() string {
...
+		if _, ok := supportedKeywordsMap()[keyword]; ok {
```
zzjin, line 112: Do not call func every time under for loop.

zzjin (approved):

PR #2293, opened 2022-12-22
fix:fix queryPrice error

xiao-jay:
Signed-off-by: 晓杰 <2561589453@qq.com>

change type to volumn-type in queryPrice

### DIFF
--- a/controllers/infra/api/v1/infra_types.go
+++ b/controllers/infra/api/v1/infra_types.go
@@ -254,23 +254,27 @@ func (i *Infra) GetInstancesAndVolumesTag() string {
 }
 
 // QueryPrice query infra price/hour, unit: CNY cents/hour
+// may be error is not nil,but the price should calculate
 func (i *Infra) QueryPrice() (int64, error) {
 	valueEc2, valueEbs := float64(0), float64(0)
+	var err error
 	for _, j := range i.Spec.Hosts {
 		if _, ok := ec2NorthPrice[j.Flavor]; !ok {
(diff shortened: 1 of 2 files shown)


controllers/infra/api/v1/infra_types.go
```diff
@@ -254,23 +254,27 @@ func (i *Infra) GetInstancesAndVolumesTag() string {
...
+				err = fmt.Errorf(" VolumnType:%v,no ebs type :%v", disk.VolumeType, err)
+				continue
 			}
-			valueEbs += ebs[disk.Type] * float64(disk.Capacity) * float64(j.Count)
+			valueEbs += ebs[disk.VolumeType] * float64(disk.Capacity) * float64(j.Count)
 		}
 	}
 	valueEbs = valueEbs / 30 / 24
-	return int64(valueEc2 + valueEbs), nil
+	return int64(valueEc2 + valueEbs), err
```
fanux, line 277: So, how to handle this error when call this function?

xiao-jay: in metering will only print error and not return
```
count, err := infra.QueryPrice()
r.Logger.Info("get infra :", "name: ", infra.Name, "namespace: ", infra.Namespace, "price: ", count, "meteringName", metering.Name, "userNamespace", metering.Spec.Namespace)
if err != nil {
	r.Logger.Error(err, "Failed to get InfraPrice,meteringName", metering.Name)
}
amount += count
```

fanux (approved):

Issue #2189, opened 2022-12-05
BUG:  when delete master , get apiserver config found etcd address is not sync newset config yet.


cuisongliu:
### Sealos Version

7045a3b

### Operating environment

```markdown
- Sealos version:
- Docker version:
- Kubernetes version:
- Operating system:
- Runtime environment:
- Cluster size:
- Additional information:
```

fanux:
Add more details?

cuisongliu:
when delete master , get apiserver config  found etcd address is not sync newset config yet.

cuisongliu:
https://github.com/labring/sealos/blob/88c1746ba7b9a477afc792f8210e6b3b704db363/pkg/runtime/kubeadm.go#L380 
127.0.0.1 is better ?

PR #2323, opened 2022-12-30
feature(main): fix etcd config

### LINKED ISSUE #2322: BUG: The join master's etcd address is incorrectly configured
### Sealos Version

v4.x

### How to reproduce the bug?

1 : run `sealos run kubernetes:xxxx`
2: `sealos join master xxxxxx`

This will cause the etcd-server configuration address of the subsequent master apiserver to be the first master node, which should be the ip address of all etcd nodes or the local etcd node ip (if there is an etcd on the local node). This problem can cause the cluster to fail if the first etcd fails, even if there are two healthy etcds left.

example:
```
cat /etc/kubernetes/manifests/kube-apiserver.yaml |grep etcd-servers
    - --etcd-servers=https://10.2.8.53:2379,https://10.2.8.54:2379,https://10.2.8.52:2379
    - --etcd-servers-overrides=/events#https://10.2.8.53:2381;https://10.2.8.54:2381;https://10.2.8.52:2381
```

### What do you see instead?

_No response_

### Operating environment

```markdown
- Sealos version:
- Docker version:
- Kubernetes version:
- Operating system:
- Runtime environment:
- Cluster size:
- Additional information:
```

### Additional information

_No response_

cuisongliu:
close #2322 #2189

### DIFF
--- a/pkg/runtime/kubeadm.go
+++ b/pkg/runtime/kubeadm.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"path"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/labring/sealos/pkg/constants"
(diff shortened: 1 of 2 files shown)


cuisongliu:
master0:
```
root@default-node-0:~# cat /etc/kubernetes/manifests/kube-apiserver.yaml |grep etcd-servers
    - --etcd-servers=https://127.0.0.1:2379
```


master1:
```
root@default-node-1:~/.sealos/default/etc# cat /etc/kubernetes/manifests/kube-apiserver.yaml |grep etcd-servers
    - --etcd-servers=https://127.0.0.1:2379
```

fanux (approved):

PR #2339, opened 2023-01-04
feature: upgrade cluster

mond77:

### DIFF
--- a/pkg/runtime/upgrade.go
+++ b/pkg/runtime/upgrade.go
@@ -77,9 +77,10 @@ func (k *KubeadmRuntime) upgradeNode(ips []string, versoin string) error {
 		now := time.Now()
 		//restart api-server takes a few seconds
 		for err != nil {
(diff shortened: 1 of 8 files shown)


pkg/runtime/upgrade.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			return err
+		}
+		err = k.sshCmdAsync(ip,
+			//install kubeadm:{version} at the node
+			fmt.Sprintf("cp -rf %s/kubeadm /usr/bin", kubeBinaryPath),
+			//upgrade other control-plane and nodes
+			upradeNodeCmd,
+			//kubectl drain <node-to-drain> --ignore-daemonsets
+			fmt.Sprintf(drainNodeCmd, nodename),
```
oldthreefeng, line 64: drain node is dangerous ... recommend not to do this.

mond77: replace with cordon in new PR

PR #2402, opened 2023-01-13
fix: modify delete logic to support cloud provider

xiaohan1202:
bug: Cloud provider will firstly delete instance and then the control-plane will lost connection with the host, which causes deleting node in k8s failed.
solution:  Delete the node in k8s first and then reset the node, this operating order would prevent deleting failure when the nodes to be deleted crashed or not reachable.

### DIFF
--- a/pkg/apply/processor/scale.go
+++ b/pkg/apply/processor/scale.go
@@ -150,8 +150,8 @@ func (c *ScaleProcessor) DeleteCheck(cluster *v2.Cluster) error {
 	logger.Info("Executing pipeline DeleteCheck in ScaleProcessor.")
 	var ips []string
 	ips = append(ips, cluster.GetMaster0IPAndPort())
-	ips = append(ips, c.MastersToDelete...)
-	ips = append(ips, c.NodesToDelete...)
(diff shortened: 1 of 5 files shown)


pkg/apply/processor/scale.go
```diff
@@ -150,8 +150,8 @@ func (c *ScaleProcessor) DeleteCheck(cluster *v2.Cluster) error {
 	logger.Info("Executing pipeline DeleteCheck in ScaleProcessor.")
 	var ips []string
 	ips = append(ips, cluster.GetMaster0IPAndPort())
-	ips = append(ips, c.MastersToDelete...)
-	ips = append(ips, c.NodesToDelete...)
+	//ips = append(ips, c.MastersToDelete...)
```
fanux, line 153: Will deleting this cause some problems？ such as on bare metal

xiaohan1202: Actually i don't know the meaning of this checking exactly,  in my view, ensuring all these ssh-connect is simply used for cleaning the nodes to be deleted, which is optional considering cloud provider, but i'm not sure..

fanux (approved):

PR #2443, opened 2023-01-20
fix(auth): replace user id with nanoId

Abingcbc:

### DIFF
--- a/pkg/auth/utils/config.go
+++ b/pkg/auth/utils/config.go
@@ -19,17 +19,22 @@ import (
 	"fmt"
 	"time"
 
-	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
 
+	"github.com/jaevor/go-nanoid"
 	"k8s.io/apimachinery/pkg/watch"
 
-	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/labring/sealos/pkg/auth/conf"
 	"github.com/labring/sealos/pkg/client-go/kubernetes"
+)
+
+const (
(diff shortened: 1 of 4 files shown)


pkg/auth/utils/config.go
```diff
@@ -46,32 +51,38 @@ func CreateOrUpdateKubeConfig(uid string) error {
...
 	})
-	user, err := resource.Get(context.Background(), uid, metav1.GetOptions{})
-	if k8sErrors.IsNotFound(err) {
-		_, err := resource.Create(context.TODO(), &unstructured.Unstructured{Object: map[string]interface{}{
+	labelSelector, _ := labels.Parse("uid=" + uid)
+	users, err := resource.List(context.Background(), metav1.ListOptions{LabelSelector: labelSelector.String()})
+	if err != nil {
+		return err
+	}
+	if len(users.Items) == 0 {
```
zzjin, line 59: Should we check len(users.Items) must be 1?

pkg/auth/utils/config.go
```diff
@@ -46,32 +51,38 @@ func CreateOrUpdateKubeConfig(uid string) error {
...
-	} else if err != nil {
-		return err
 	} else {
 		// To trigger controller reconcile
-		user.Object["metadata"].(map[string]interface{})["labels"].(map[string]interface{})["updateTime"] = time.Now().Format("T2006-01-02T15-04-05")
-		_, err = resource.Update(context.Background(), user, metav1.UpdateOptions{})
-		if err != nil {
-			return err
+		for _, user := range users.Items {
+			user.Object["metadata"].(map[string]interface{})["labels"].(map[string]interface{})["updateTime"] = time.Now().Format("T2006-01-02T15-04-05")
```
cuisongliu, line 81: The writing is a bit complicated
```
labels:=user.GetLabels()[updateTime]=xxx
user.SetLabels(label)
```

zzjin (changes requested):
Need to add old users sync script.

fanux:
@zzjin I accept to clean up the old users directly, the migration cost is too high, and such things will not happen in the future

zzjin (approved):

zzjin (approved):

PR #2463, opened 2023-01-30
fixed issues with creating infra

zjy365:
Signed-off-by: zhu jing yang <3161362058@qq.com>


zjy365:
fixed the user balance problem

PR #2464, opened 2023-01-30
checker draft part1

cdjianghan:
this is sealos precheck checkers draft !!! 
this is sealos precheck checkers draft !!! 
this is sealos precheck checkers draft !!! 

this is part of sealos precheck implement .
main logic was implement in precheck.go and different checkers.

what i did ?
* add precheck door in pkg/runtime. why this place ? only this place can obtain dynamic port hostname,cfg etc.
* change checker interface , check() func return two values ,the first is an array of warnings ,the second is an array of errors.with this format ,when checker checks ,it won't interrupt by something wrong happened.
* add cpu , mem ,ip , net ,ssh ,kubeadm ,and some other checkers.
* the total check phase was divided into three phase ,the first is cri check(not implement now),the second is kubeadm init check ,the third is kubeadm join checks.
* I use a lot ssh to replace code logic in checker ,the reason is that this may be best way to support multipy node checks.

### DIFF
--- a/pkg/checker/check_list.go
+++ b/pkg/checker/check_list.go
@@ -15,7 +15,10 @@
 package checker
 
 import (
+	"bytes"
 	"fmt"
+	"io"
+	"os"
 
 	v2 "github.com/labring/sealos/pkg/types/v1beta1"
 )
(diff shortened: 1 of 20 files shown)


pkg/checker/check_list.go
```diff
@@ -27,14 +30,30 @@ const (
...
+	// Check checks the cluster status,which will return two value list, the first is warnings, the second is error.
+	Check(cluster *v2.Cluster, phase string) (warnings, errorList []error)
 }
 
 func RunCheckList(list []Interface, cluster *v2.Cluster, phase string) error {
-	for _, l := range list {
-		if err := l.Check(cluster, phase); err != nil {
-			return fmt.Errorf("failed to run checker: %v", err)
+	var errsBuffer bytes.Buffer
+	for _, c := range list {
```
fanux, line 40: You can add a feature that ignores a certain check

Issue #2471, opened 2023-01-31
BUG: sealos cloud cluster cant sealos run app


cdjianghan:
### Sealos Version

v4.1.4 maybe，I just click the button on desktop

### How to reproduce the bug?

after sealos cloud provider build a new cluster , I run a new app ,such as sealos run openebs or helm ,
![84891675178218_ pic](https://user-images.githubusercontent.com/52602064/215818719-ffc9b0da-b2a6-4cd4-878a-0ef3d36fb515.jpg)
there is a Clusterfile under path .sealos/default/   ,but the cluster file has ts suffix.
after I delete the suffix ,sealos run is ok.
![85101675179666_ pic](https://user-images.githubusercontent.com/52602064/215819305-fc7ca8ee-b7b3-45bc-bf60-ba4428ba5a1f.jpg)

### Operating environment

```markdown
- Sealos version:
- Docker version:
- Kubernetes version:
- Operating system:
- Runtime environment:
- Cluster size:
- Additional information:
```

PR #2476, opened 2023-02-01
fix: configure default clustername

xiaohan1202:
fix issue: https://github.com/labring/sealos/issues/2471

### DIFF
--- a/controllers/infra/testdata/api/infra.go
+++ b/controllers/infra/testdata/api/infra.go
@@ -19,7 +19,7 @@ spec:
   - roles: [master] 
     count: 1
     flavor: t2.medium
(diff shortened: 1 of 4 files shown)


controllers/infra/testdata/api/infra.go
```diff
@@ -19,7 +19,7 @@ spec:
   - roles: [master] 
     count: 1
     flavor: t2.medium
-    image: "ami-0d66b970b9f16f1f5"
+    image: "ami-09123565dfe16d662"
```
zzjin, line 22: 时不时需要前端统一同步更新?

fanux (approved):

Issue #2496, opened 2023-02-04
BUG: cluster status is not ClusterSuccess


willzhang:
### Sealos Version

v4.1.5-alpha2

### How to reproduce the bug?

install first app failed
```
root@node31:~# sealos run registry.cn-shenzhen.aliyuncs.com/cnmirror/eck-operator:v2.6.1
2023-02-04T13:48:17 info start to install app in this cluster
Trying to pull registry.cn-shenzhen.aliyuncs.com/cnmirror/eck-operator:v2.6.1...
Getting image source signatures
Copying blob 49c8d6358870 done  
Copying config 2d48c36d7f done  
Writing manifest to image destination
Storing signatures
2023-02-04T13:51:56 info Executing pipeline MountRootfs in InstallProcessor.
2023-02-04T13:51:57 info Executing pipeline MirrorRegistry in InstallProcessor.
2023-02-04T13:51:59 info Executing UpgradeIfNeed Pipeline in InstallProcessor
Release "eck-operator" does not exist. Installing it now.
Error: rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition "agents.agent.k8s.elastic.co" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "eck-operator": current value is "eck"; annotation validation error: key "meta.helm.sh/release-namespace" must equal "elastic-system": current value is "default"
2023-02-04T13:52:00 error Applied to cluster error: exit status 1
Error: exit status 1
exit status 1
```

install again or other app can not install to cluster
```
root@node31:~# sealos run docker.io/labring/nginx:v1.23.3
Error: cluster status is not ClusterSuccess
cluster status is not ClusterSuccess
```

temporary solution
```
sed -i "s#phase: ClusterFailed#phase: ClusterSuccess#g" /root/.sealos/default/Clusterfile
```

### What is the expected behavior?

install app failed do not let cluster status failed.

### What do you see instead?

install app failed  cluster status will be failed.

### Operating environment

```markdown
- Sealos version:
- Docker version:
- Kubernetes version:
- Operating system:
- Runtime environment:
- Cluster size:
- Additional information:
```

### Additional information

maybe should mark `app_status` not `cluster_status` when app install failed.

and the phase option should in cmd level

```
  - cmd:
    - helm upgrade -i nginx charts/nginx -n nginx --create-namespace
      phase: AppSuccess
```

not
```
  - cmd:
[...]

PR #2505, opened 2023-02-06
fix: cluster status

xiaohan1202:
fix issue: https://github.com/labring/sealos/issues/2496

### DIFF
--- a/pkg/apply/applydrivers/apply_drivers_default.go
+++ b/pkg/apply/applydrivers/apply_drivers_default.go
@@ -76,6 +76,7 @@ type Applier struct {
 func (c *Applier) Apply() error {
 	clusterPath := constants.Clusterfile(c.ClusterDesired.Name)
 	var err error
+	var errCode constants.ErrorCode
 	defer func() {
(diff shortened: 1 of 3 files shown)


pkg/apply/applydrivers/apply_drivers_default.go
```diff
@@ -112,7 +113,7 @@ func (c *Applier) initStatus() {
 
 // todo: atomic updating status after each installation for better reconcile?
 // todo: set up signal handler
-func (c *Applier) updateStatus(err error) {
+func (c *Applier) updateStatus(errCode constants.ErrorCode, err error) {
```
zzjin, line 116: Return two error, one for app apply, another for cluster update is better.
As in golang, error can be any type that implmented   `Error()`, Not return custom type, but return error type with custom errors.
Like:
```golang
var ErrNotMatch = errors.New("NotMatch")

func do() error {
    return ErrNotMatch
}

if errors.Is(err, ErrNotMatch) {
    // RODO
}
```

zzjin (changes requested):
Need update

PR #2511, opened 2023-02-06
add kubernetes version cri version output in sealos verison cmd

cdjianghan:


pkg/version/version.go
```diff
@@ -35,3 +43,35 @@ func Get() Info {
 		Platform:   fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH),
 	}
 }
+
+func GetKubernetesVersion(cluster *v2.Cluster) *KubernetesVersion {
```
zzjin, line 47:
```suggestion
func GetKubernetesVersion(cluster *v2.Cluster) (*KubernetesVersion, error) {
```
return error if cmd exec fail or version info is mismatch when set debug=true,and print it out.

Do not omit any errors.

fengxsong: > return error if cmd exec fail or version info is mismatch when set debug=true,and print it out.
> 
> Do not omit any errors.

but what if `kubectl` or `crictl` binary are missing?

zzjin: After discussion, will show not found and suggest user to start kubernetes cluster using `sealos run`.
image

pkg/version/types.go
```diff
@@ -16,21 +16,61 @@ limitations under the License.
...
-	BuildDate  string `json:"buildDate"`
-	GoVersion  string `json:"goVersion"`
-	Compiler   string `json:"compiler"`
-	Platform   string `json:"platform"`
+	GitVersion string `json:"gitVersion" yaml:"gitVersion"`
+	GitCommit  string `json:"gitCommit,omitempty" yaml:"gitCommit,omitempty"`
+	BuildDate  string `json:"buildDate" yaml:"buildDate"`
+	GoVersion  string `json:"goVersion" yaml:"goVersion"`
+	Compiler   string `json:"compiler" yaml:"compiler"`
+	Platform   string `json:"platform" yaml:"platform"`
```
lingdie, line 32: why add yaml?

cdjianghan: output can use “-o yaml | json” to define output format

lingdie: But in k8s crd, we can '-o yaml' to format output too, and we don't re-define `yaml` filed, what's the difference?

cdjianghan: no difference ,just for format convert convience.
another point is crictl dont have -o options

zzjin (approved):

PR #2518, opened 2023-02-07
feat. add repo public for hubauth svc.

lingdie:


service/hub/server/config.go
```diff
@@ -82,6 +102,8 @@ func LoadConfig(fileName string) (*Config, error) {
 	if err = validate(c); err != nil {
 		return nil, fmt.Errorf("invalid config: %s", err)
 	}
+	// store client globally
+	k8sClient, _ = kubernetes.NewKubernetesClient(c.Server.Kubeconfig, "")
```
lingdie, line 106: split cheaker and var initer.

service/hub/server/config.go
```diff
@@ -8,10 +8,23 @@ import (
...
+	"github.com/cesanta/glog"
 	"github.com/docker/libtrust"
+	"github.com/labring/sealos/pkg/client-go/kubernetes"
 	yaml "gopkg.in/yaml.v2"
+	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+var k8sClient kubernetes.Client
+
+func init() {
```
lingdie, line 20: why create var when first use anonymous

service/hub/server/config.go
zzjin: Why reassign global k8sClient to new user spec kubeconfig?

lingdie: deleted.

service/hub/server/config.go
```diff
@@ -98,5 +102,10 @@ func LoadConfig(fileName string) (*Config, error) {
 	if !tokenConfigured {
 		return nil, fmt.Errorf("failed to load token cert and key: none provided")
 	}
+	// setup k8sClient by using controller-runtime
+	k8sClient, err = kubernetes.NewKubernetesClientByConfig(ctrl.GetConfigOrDie())
```
zzjin, line 106: init here, but k8sClient never called?

zzjin (approved):

PR #2519, opened 2023-02-07
bug: optimize installApp

lingdie:
optimize 
- save write back file: fix codes
- install new image: add log and fix image pull error does not return

### DIFF
--- a/pkg/apply/applydrivers/apply_drivers_default.go
+++ b/pkg/apply/applydrivers/apply_drivers_default.go
@@ -76,11 +76,13 @@ type Applier struct {
 func (c *Applier) Apply() error {
 	clusterPath := constants.Clusterfile(c.ClusterDesired.Name)
 	var clusterErr, appErr error
+	// save cluster to file after apply
 	defer func() {
 		logger.Debug("write cluster file to local storage: %s", clusterPath)
-		saveerror := yaml.MarshalYamlToFile(clusterPath, c.getWriteBackObjects()...)
-		if clusterErr == nil {
-			clusterErr = saveerror
+		saveErr := yaml.MarshalYamlToFile(clusterPath, c.getWriteBackObjects()...)
+		if saveErr != nil {
+			logger.Error("write cluster file to local storage: %s error, %s", clusterPath, saveErr)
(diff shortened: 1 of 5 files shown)


pkg/apply/processor/install.go
```diff
@@ -117,7 +100,28 @@ func (c *InstallProcessor) SyncStatusAndCheck(cluster *v2.Cluster) error {
...
+	}
+
+	prompt := fmt.Sprintf("are you sure to override these following apps? \n%s\t", strings.Join(c.imagesToOverride, "\n"))
+	cancel := "you have canceled to override these apps!"
+	pass, err := confirm.Confirm(prompt, cancel)
+	if err != nil {
+		return err
+	}
+	if !pass {
+		return errors.New(cancel)
```
fengxsong, line 117: should ignore cancel error?

pkg/apply/applydrivers/apply_drivers_default.go
```diff
@@ -92,6 +95,11 @@ func (c *Applier) Apply() error {
 		c.ClusterDesired.CreationTimestamp = c.ClusterCurrent.CreationTimestamp
 	}
 	c.updateStatus(clusterErr, appErr)
```
fengxsong, line 97: `appErr` is not used in `updateStatus`

pkg/apply/processor/install.go
lingdie: Only moved code place. Do you mean we should ignore appErr if it's cancel?

pkg/apply/applydrivers/apply_drivers_default.go
lingdie: I will change func updateStatus and will use appErr to generate cmdPhase for each cmd.

pkg/apply/processor/install.go
fengxsong: > Only moved code place. Do you mean we should ignore appErr if it's cancel?

yes, cancel is not an unexpected error.

lingdie: ok, add a check at return appErr.

cuisongliu (approved):

PR #2545, opened 2023-02-10
fix sealos version err return bug

cdjianghan:
fix sealos version error return bug


cuisongliu (approved):

cuisongliu:
https://github.com/labring-actions/bug-verfiy/actions/runs/4140868401/jobs/7159930134

PR #2551, opened 2023-02-10
fix: inspect function support resolving shortname

### LINKED ISSUE #2524: BUG: 镜像load后名字为localhost
### Sealos Version

v4.1.4 v4.1.5-alpha2

### How to reproduce the bug?

image
执行命令为 :
sealos pull labring/kubernetes:v1.24.9
sealos save -o kubernetes.tar labring/kubernetes:v1.24.9
sealos load -i kubernetes.tar

load进去的镜像会变成localhost

_No response_

### What do you see instead?

_No response_

### Operating environment

```markdown
- Sealos version:
- Docker version:
- Kubernetes version:
- Operating system:
- Runtime environment:
- Cluster size:
- Additional information:
```

### Additional information

_No response_

24sama:
Fix #2524 

Currently, there is an error when executing the run command with a short name:

image

This PR adds the `localhost` and `docker.io` prefixes to the shortname.

image

More info about the shortname, please see: https://github.com/containers/buildah/issues/1034

### DIFF
--- a/pkg/buildah/inspect.go
+++ b/pkg/buildah/inspect.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/containers/common/pkg/retry"
 	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/pkg/shortnames"
 	"github.com/opencontainers/go-digest"
 
 	"github.com/containers/buildah"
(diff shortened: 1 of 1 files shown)


zzjin:
Thank you for your first contribution!

pkg/buildah/inspect.go
```diff
@@ -250,16 +252,42 @@ func inspectImage(ctx context.Context, sc *types.SystemContext, store storage.St
...
-	ref, err := transport.ParseReference(imgName)
+
+	// try out the candidates as resolved by shortnames. This takes
+	// "localhost/" prefixed images into account as well.
+	candidates, err := shortnames.ResolveLocally(sc, imgName)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("error resolve shortname %s: %w", imgName, storage.ErrImageUnknown)
+	}
+	// also try out the "docker.io/" prefixed image
```
zzjin, line 262: I agree with the `ResolveLocally` functional, but seems we do not need to auto parse imgName for `docker.io` with two reasons below:
1. If the image belongs to `quay.io` or `ghcr.io`
2. if user is using one mirror registry like `dockerproxy.com/docker.io/`

zzjin:
@fengxsong Is this pr also resolves #2388 ?

fanux:
> 

Yes~

pkg/buildah/inspect.go
24sama: Thanks for your review! This function `ParseDockerRef()` is just for backward compatibility. the function returns digested reference, e.g. `docker.io/library/busybox:latest@sha256:xxxx` will be returned as `docker.io/library/busybox@sha256:xxxx`.

Indeed, I guess we didn't have this scenario, so I have removed this function.

cuisongliu (approved):

PR #2566, opened 2023-02-13
feat: add Failed status for cloud provider

xiaohan1202:

### DIFF
--- a/controllers/cluster/api/v1/cluster_types.go
+++ b/controllers/cluster/api/v1/cluster_types.go
@@ -82,6 +82,7 @@ type ClusterStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.status",description="Status of Cluster in group"
 
 // Cluster is the Schema for the clusters API
 type Cluster struct {
--- a/controllers/infra/controllers/infra_controller.go
+++ b/controllers/infra/controllers/infra_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	v1 "github.com/labring/sealos/controllers/cluster/api/v1"
 	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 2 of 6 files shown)


controllers/cluster/api/v1/cluster_types.go
```diff
@@ -82,6 +82,7 @@ type ClusterStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.status",description="Status of Cluster in group"
```
cuisongliu, line 85: Phase is better ?

controllers/infra/controllers/infra_controller.go
```diff
@@ -251,6 +269,44 @@ func getSecretName(infra *infrav1.Infra) string {
...
+			return err
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *InfraReconciler) updateFinalizers(ctx context.Context, objectKey client.ObjectKey, obj client.Object, finalizers []string) error {
```
cuisongliu, line 289: see user controller usage.

controllers/cluster/api/v1/cluster_types.go
xiaohan1202: this format is the same as infra CRD

controllers/infra/controllers/infra_controller.go
```diff
@@ -91,15 +92,22 @@ func (r *InfraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		return ctrl.Result{}, err
 	}
 
+	// if infra status is failed, delete finalizer and do not reconcile
+	if infra.Status.Status == v1.Failed.String() {
```
cuisongliu, line 96:
```
if _, err := r.finalizer.RemoveFinalizer(ctx, infra, r.DeleteInfra); err != nil {
			return ctrl.Result{}, err
		}
		return ctrl.Result{}, nil
```

cuisongliu (approved):

Issue #2562, opened 2023-02-13
Proposal: predefine a directory to store configuration file like kubeadm, kubelet configs


fengxsong:
### Create the respective sealos cluster image distribution

1. predefine a directory name, like `init-configs` or whatever, in sealos's code base;
2. place customized kubeadm/kubelet/kubeproxy config files in it, and build a customized cluster image from the official image.
3. run `sealos apply` with the newly built cluster image without providing kubernetes's related config files, `sealos` will automatically merge those files into the default kubeadmConfigs.

fengxsong:
cc @cuisongliu

cuisongliu:
That's right, I think the directory is kubeadm or directly defined in the etc/kubeadm.yaml configuration.

PR #2567, opened 2023-02-13
refactor: adjust the order of merging kubeadm configs

fengxsong:

Seems it's already implemented but needs some adjustments. #2562

### DIFF
--- a/pkg/runtime/kubeadm.go
+++ b/pkg/runtime/kubeadm.go
@@ -113,14 +113,20 @@ func (k *KubeadmRuntime) getCGroupDriver(node string) (string, error) {
 }
 
 func (k *KubeadmRuntime) MergeKubeadmConfig() error {
+	k.ImageKubeVersion = k.getKubeVersionFromImage()
+	for _, fn := range []string{
+		"",                          // generate default kubeadm configs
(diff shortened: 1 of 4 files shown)


pkg/runtime/kubeadm.go
```diff
@@ -113,15 +113,21 @@ func (k *KubeadmRuntime) getCGroupDriver(node string) (string, error) {
 }
 
 func (k *KubeadmRuntime) MergeKubeadmConfig() error {
+	k.ImageKubeVersion = k.getKubeVersionFromImage()
+	for _, fn := range []string{
+		"",                          // generate default kubeadm configs
+		k.getDefaultKubeadmConfig(), // merging from predefined path of file if file exists
```
cuisongliu, line 119: 1. k.getDefaultKubeadmConfig()
2. “”

is OK

fengxsong: This way in the comment, as well as the original, both are setting default values first, then dst will not be overwritten even if the src is set. IMO, with mergo's transformers is more intuitive.

cuisongliu:
https://github.com/labring-actions/bug-verify/actions/runs/4162757941 test is ok

cuisongliu (approved):

PR #2578, opened 2023-02-14
CI: add test-cmd ci.

lingdie:


xiao-jay:
you can also test sealos apply、build、run、reset、pull、push、login in ci

cuisongliu:
https://github.com/labring-actions/bug-verify/blob/main/.github/workflows/inspect.yaml  example is there

PR #2583, opened 2023-02-14
fix: issue #2577

### LINKED ISSUE #2577: BUG: inspect bug
### Sealos Version

main

### How to reproduce the bug?

cuisongliu@cuisongliu:~/GolandProjects/src/github.com/labring/sealos$ bin/linux_amd64/sealos docker-archive:/home/cuisongliu/Downloads/qps_test_0_4_2.tar 
Error: unknown command "docker-archive:/home/cuisongliu/Downloads/qps_test_0_4_2.tar" for "sealos"
Run 'sealos --help' for usage.
cuisongliu@cuisongliu:~/GolandProjects/src/github.com/labring/sealos$ bin/linux_amd64/sealos inspect  docker-archive:/home/cuisongliu/Downloads/qps_test_0_4_2.tar 
Error: error resolve shortname /home/cuisongliu/Downloads/qps_test_0_4_2.tar: image not known
cuisongliu@cuisongliu:~/GolandProjects/src/github.com/labring/sealos$ 

_No response_

### What do you see instead?

_No response_

### Operating environment

```markdown
- Sealos version:
- Docker version:
- Kubernetes version:
- Operating system:
- Runtime environment:
- Cluster size:
- Additional information:
```

### Additional information

_No response_

fengxsong:

fix #2577

### DIFF
--- a/pkg/buildah/inspect.go
+++ b/pkg/buildah/inspect.go
@@ -25,7 +25,6 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/containers/common/pkg/retry"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/shortnames"
 	"github.com/opencontainers/go-digest"
@@ -67,6 +66,8 @@ func (opts *inspectResults) RegisterFlags(fs *pflag.FlagSet) {
 	fs.SetInterspersed(false)
 	fs.StringVarP(&opts.format, "format", "f", "", "use `format` as a Go template to format the output")
 	fs.StringVarP(&opts.inspectType, "type", "t", inspectTypeContainer, "look at the item of the specified `type` (container or image) and name")
+	fs.Bool("tls-verify", false, "require HTTPS and verify certificates when talking to the container registry or daemon")
+	bailOnError(markFlagsHidden(fs, "tls-verify"), "")
 }
 
 func newInspectCommand() *cobra.Command {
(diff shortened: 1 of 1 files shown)


pkg/buildah/inspect.go
```diff
@@ -252,37 +248,34 @@ func inspectImage(ctx context.Context, sc *types.SystemContext, store storage.St
...
-			logger.Debug("using image name %q", candidate.String())
-			src = res
-			break
+			return nil, nil, fmt.Errorf("error resolve shortname %s: %w", imgName, storage.ErrImageUnknown)
+		}
+		for _, name := range names {
+			src, err = parseSource(name.String())
+			if err == nil {
+				break
+			}
```
24sama, line 272: if the second `err` have a value, we can not handle it.

fengxsong: I don't understand what you are saying this is a bug. 🤔️

24sama, line 272: Initially, an `err` (we can call it err1) variable is declared via `var`, and this line `names, err := shortnames.ResolveLocally(sc, imgName)` (we can call it err2) is also declare an `err` variable, which is in the `case TransportContainersStorage:`. 

So, if `src, err = parseSource(name.String())` return an error, it will assign the value to err2, which will eventually cause the function to exit without returning an error and the `src` variable is still a nil pointer.

fengxsong: I see, thanks for your kind interpretation. 👍

Issue #2317, opened 2022-12-29
Feature: sealos run --single should auto untaint master nodes


willzhang:
### What is the problem this feature will solve?

after `sealos run --single` should do this everytime manaul
```
root@ubuntu:~# kubectl taint nodes --all node-role.kubernetes.io/master:NoSchedule-
node/ubuntu untainted
```

### If you have solution，please describe it

if sealos run --single

then
```
kubeadm init .... --skip-phases=mark-control-plane

---
apiVersion: kubeadm.k8s.io/v1beta3
kind: InitConfiguration
skipPhases:
- mark-control-plane
```
or
```
---
apiVersion: kubeadm.k8s.io/v1beta3
kind: InitConfiguration
nodeRegistration:
  taints:
  - effect: PreferNoSchedule
    key: node-role.kubernetes.io/master
```

PR #2654, opened 2023-02-23
feature: single host should allow scheduling to master

hexinatgithub:
Every time I run a Single host cluster, I need to manual remove the Taint.

```shell
$ sealos run labring/kubernetes:v1.25.0 labring/helm:v3.8.2 labring/calico:v3.24.1 --single
# remove taint
$ kubectl taint nodes --all node-role.kubernetes.io/master:NoSchedule-
```

I add a patch to do the untaint automatic #2317.

### DIFF
--- a/.github/workflows/test_run.yml
+++ b/.github/workflows/test_run.yml
@@ -35,6 +35,53 @@ jobs:
         with:
           name: sealos
           path: /usr/bin/sealos
+
+  verify-run-single:
+    needs: [build-sealos]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download sealos
+        uses: actions/download-artifact@v3
+        with:
+          name: sealos
+          path: /tmp/
+      - name: Verify sealos
+        run: |
+          sudo chmod a+x /tmp/sealos
+          sudo mv /tmp/sealos /usr/bin/
+          sudo sealos version
+      - name: Remove containerd && docker
+        uses: labring/sealos-action@v0.0.7
+        with:
+          type: prune
+      - name: Auto install k8s using sealos
+        run: |
+          sudo sealos run labring/kubernetes:v1.25.0 --single --debug
+          mkdir -p "$HOME/.kube"
+          sudo cp -i /etc/kubernetes/admin.conf "$HOME/.kube/config"
+          sudo chown "$(whoami)" "$HOME/.kube/config"
+          kubectl get svc
+          kubectl get pod -A
+          sudo cat /root/.sealos/default/etc/kubeadm-init.yaml
+          sudo cat /root/.sealos/default/Clusterfile
+      - name: Verify Single Node Cluster Status
+        run: |
+          echo "Verify Cluster"
+          echo "Current system info"
+          sudo /var/lib/sealos/data/default/rootfs/opt/sealctl cri socket
+          sudo /var/lib/sealos/data/default/rootfs/opt/sealctl cri cgroup-driver --short
+          echo "Check Single Node Stains"
+          taints=$(kubectl get nodes -o json | jq ".items[].spec.taints")
+          if [ grep -q "control-plane"< <(echo $taints) ]; then
+            printf "node taints: %s\n" $taints && false
+          fi
+          echo "Current Cluster info"
+          set -e
+          sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep /run/containerd/containerd.sock
+          sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep systemd
+          sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep 100.64.0.0/10
+          sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep 10.96.0.0/22
+
   verify-run-containerd:
     needs: [build-sealos]
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


cuisongliu:
```
kubeadm init .... --skip-phases=mark-control-plane

---
apiVersion: kubeadm.k8s.io/v1beta3
kind: InitConfiguration
skipPhases:
- mark-control-plane
```

hexinatgithub:
That's sounds great, I will change my implementation.

cuisongliu:
test is ok ?

hexinatgithub:
> test is ok ?

Test tomorrow, may contain some bugs.

hexinatgithub:
Some test case need to add, need to improve.

cuisongliu:
You can refer to the test case in ci and test it with the shell to see if the final single node has stains.

cuisongliu:
https://github.com/labring/sealos/actions/runs/4259381988/jobs/7412043583#step:6:243 same not remove taint ?

fengxsong:
this process should only invoke in `initCluster`, so I would rather implement it in function `pkg/runtime/KubeadmRuntime.generateInitConfigs`. 

```
if sum(len(masters), len(nodes)) == 1 then append skip-phases
```

cuisongliu:
> this process should only invoke in `initCluster`, so I would rather implement it in function `pkg/runtime/KubeadmRuntime.generateInitConfigs`.
> 
> ```
> if sum(len(masters), len(nodes)) == 1 then append skip-phases
> ```

If he just installs the join node behind master0, is this judgment a bit rough?

hexinatgithub:
> 

I think dependency injection will more appropriate

hexinatgithub:
[NewDefaultRuntime](https://github.com/hexinatgithub/sealos/blob/b05ff82d702e9ae757f422b8ff622233a9f25c86/pkg/runtime/runtime.go#L126) is a really confused function signature which provide `*KubeadmConfig` but will not use it.

fengxsong:
> > this process should only invoke in `initCluster`, so I would rather implement it in function `pkg/runtime/KubeadmRuntime.generateInitConfigs`.
> > ```
> > if sum(len(masters), len(nodes)) == 1 then append skip-phases
> > ```
> 
> If he just installs the join node behind master0, is this judgment a bit rough?

once cluster is provisioned, the taints should remove manually.

hexinatgithub:
> > > this process should only invoke in `initCluster`, so I would rather implement it in function `pkg/runtime/KubeadmRuntime.generateInitConfigs`.
> > > ```
> > > if sum(len(masters), len(nodes)) == 1 then append skip-phases
> > > ```
> > 
> > 
> > If he just installs the join node behind master0, is this judgment a bit rough?
> 
> once cluster is provisioned, the taints should remove manually.

It should only skip `mark-control-plane` phase when the cluster is first created and `--single` flag is set.
When user need join more nodes into cluster, he shouldn't provide `--single` flag the first time, the Taint should remain.

cuisongliu:
https://github.com/labring/sealos/actions/runs/4260038410/jobs/7413126020#step:6:250 There should be no stains

hexinatgithub:
> https://github.com/labring/sealos/actions/runs/4260038410/jobs/7413126020#step:6:250 There should be no stains

(CreateProcessor.PreProcess)[https://github.com/hexinatgithub/sealos/blob/b05ff82d702e9ae757f422b8ff622233a9f25c86/pkg/apply/processor/create.go#L87] use [NewDefaultRuntime](https://github.com/hexinatgithub/sealos/blob/b05ff82d702e9ae757f422b8ff622233a9f25c86/pkg/runtime/runtime.go#L126) cause this weird bug.

pkg/apply/processor/create.go
```diff
@@ -89,7 +89,7 @@ func (c *CreateProcessor) PreProcess(cluster *v2.Cluster) error {
 	if err := MountClusterImages(cluster, c.Buildah); err != nil {
 		return err
 	}
-	runTime, err := runtime.NewDefaultRuntime(cluster, c.ClusterFile.GetKubeadmConfig())
```
cuisongliu, line 92: This logic needs to be merged, and this change may fail the test.

cuisongliu:
> It was accidentally turned off. Please reopen this issue.

你重新提交pr即可

hexinatgithub:
I rewrite my code and pass all test. 
Please check it.

.github/workflows/test_run.yml
```diff
@@ -35,6 +35,50 @@ jobs:
...
+          sudo cat /root/.sealos/default/etc/kubeadm-init.yaml
+          sudo cat /root/.sealos/default/Clusterfile
+      - name: Verify Single Node Cluster Status
+        run: |
+          echo "Verify Cluster"
+          echo "Current system info"
+          sudo /var/lib/sealos/data/default/rootfs/opt/sealctl cri socket
+          sudo /var/lib/sealos/data/default/rootfs/opt/sealctl cri cgroup-driver --short
+          echo "Check Single Node Stains"
+          ! kubectl get nodes -o json | jq ".items[].spec.taints" | grep "control-plane"
```
cuisongliu, line 74: It seems that there is no error reported.

pkg/clusterfile/clusterfile.go
```diff
@@ -43,6 +43,7 @@ type Interface interface {
 	GetCluster() *v2.Cluster
 	GetConfigs() []v2.Config
 	GetKubeadmConfig() *runtime.KubeadmConfig
+	SingleMode(bool) error
```
cuisongliu, line 46: SetSingleMode is better ?

cuisongliu (approved):

PR #2662, opened 2023-02-24
fix: delete local file

c121914yu:
fix: delete local file


zzjin (changes requested):
remove root dir js rela files

fanux (approved):

zzjin (approved):

PR #2666, opened 2023-02-24
fix: same issue #2623 again

fengxsong:
same issue as #2623


pkg/buildah/create.go
```diff
@@ -51,16 +51,19 @@ func newCreateCmd() *cobra.Command {
...
-				return err
-			}
 			bder, err := New("")
 			if err != nil {
 				return err
 			}
-			info, err := bder.Create(opts.name, args[0],
-				WithPlatformOption(v1.Platform{OS: oss, Architecture: arch, Variant: variant}))
+			flagSetters := []FlagSetter{}
+			if flagChanged(c, "platform") {
```
cuisongliu, line 59: Do you need to judge these parameters of arch?

cuisongliu (approved):

PR #2700, opened 2023-03-01
sealos-image-hub

zjy365:


frontend/providers/imagehub/Makefile
```diff
@@ -0,0 +1,48 @@
+
+SERVICE_NAME=zhujingyang/sealos-image-hub
```
LeezQ, line 2: `zhujingyang/sealos-image-hub`  Is the image url here incorrect? It should be placed in the public repository.

frontend/providers/imagehub/deploy/manifests/frontend.yaml
```diff
@@ -0,0 +1,96 @@
...
+      containers:
+        - name: desktop-frontend
+          securityContext:
+            runAsNonRoot: true
+            runAsUser: 1001
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - "ALL"
+          image: zhujingyang/desktop-frontend:latest
```
LeezQ, line 87: Here is the same

LeezQ (approved):

frontend/providers/imagehub/Makefile
zjy365: It will become a public warehouse later

frontend/providers/imagehub/src/layout/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+    return createSealosApp({
+      appKey: 'sealos-image-hub'
+    });
+  }, []);
+
+  useEffect(() => {
+    const initApp = async () => {
+      try {
+        const result = await sealosApp.getUserInfo();
+        setSession(result);
```
zzjin, line 22: When deploying apps along side with custom unique url domain, we do not want to store sessions under local storage for safety purpose.
Maybe getSession will just save user infos as document variables is enough.

frontend/providers/imagehub/src/stores/session.ts
```diff
@@ -0,0 +1,47 @@
...
+  setSessionProp: (key: keyof Session, value: any) => void;
+  getSession: () => Session;
+  delSession: () => void;
+  isUserLogin: () => boolean;
+  getKubeconfigToken: () => string;
+};
+
+const useSessionStore = create<SessionState>()(
+  devtools(
+    persist(
```
zzjin, line 20: Just remove this `persist` wrapper is enough./

zzjin (approved):

PR #2717, opened 2023-03-03
get terminal domain from env

gitccl:

### DIFF
--- a/controllers/terminal/controllers/terminal_controller.go
+++ b/controllers/terminal/controllers/terminal_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"os"
 	"time"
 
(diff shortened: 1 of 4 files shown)


controllers/terminal/controllers/terminal_controller.go
```diff
@@ -444,6 +446,14 @@ func buildLabelsMap(terminal *terminalv1.Terminal) map[string]string {
 	return labelsMap
 }
 
+func getDomain() string {
+	domain := os.Getenv("DOMAIN")
```
zzjin, line 450: Seems we can store env variabe `DMOAIN` at init time and just read it var at every reconceil call to save some system calls.

gitccl: done

zzjin (approved):

PR #2720, opened 2023-03-03
add new driver interface and fix some bugs

xiaohan1202:


controllers/infra/drivers/aliyun/delete_instance.go
```diff
@@ -54,86 +56,145 @@ func (d Driver) deleteInstances(hosts *v1.Hosts) error {
...
+
+	// delete instances
+	for i := range infra.Spec.Hosts {
+		hosts := infra.Spec.Hosts[i]
+		eg.Go(func() error {
+			return d.deleteInstances(&hosts)
+		})
+	}
+	if err := eg.Wait(); err != nil {
+		return fmt.Errorf("got an error deleting instances: %v", err)
```
fanux, line 123: failed to ...

xiaohan1202: > failed to ...

okay I've changed it (this log is generated by copilot, lol

controllers/infra/drivers/aliyun/create_instance.go
```diff
@@ -203,7 +203,7 @@ func getDataDisks(host *v1.Hosts) []ecs.RunInstancesDataDisk {
 }
 
 func getHostName(hosts *v1.Hosts, infra *v1.Infra) string {
-	return fmt.Sprintf("%s-%s-%d", infra.Namespace, infra.Name, hosts.Index)
+	return fmt.Sprintf("%s-%s-${instance_id}", infra.Namespace, infra.Name)
```
zzjin, line 206: What is `${instance_id}` means here?

xiaohan1202: > What is `${instance_id}` means here?

according to aliyun api docs:
image

zzjin: Use some info that we do not control is not so pretty? 
Imaging this situation: user want to know roles/disks info when only have instance name, the auto assigned ${instance_id} is not saved under infra CR, so they cannot get real instance full name?

xiaohan1202: > Use some info that we do not control is not so pretty? Imaging this situation: user want to know roles/disks info when only have instance name, the auto assigned ${instance_id} is not saved under infra CR, so they cannot get real instance full name?

User can get this instanceId from their infra cr metadata:
image

zzjin (approved):

Issue #2688, opened 2023-02-28
BUG:  custom certSANs not merged to the default ClusterConfiguration


willzhang:
### Sealos Version

v4.1.5

### Sealos Version

v4.1.5

### How to reproduce the bug?

create clusterfile with custom certSANs
```
apiVersion: apps.sealos.io/v1beta1
kind: Cluster
metadata:
  creationTimestamp: null
  name: default
spec:
  hosts:
  - ips:
    - 192.168.72.34:22
    roles:
    - master
    - amd64
  image:
  - hub.sealos.cn/labring/kubernetes:v1.25.6
  ssh:
    passwd: "123456"
    pk: /root/.ssh/id_rsa
    port: 22
status: {}

---
apiVersion: kubeadm.k8s.io/v1beta3
kind: ClusterConfiguration
apiServer:
  certSANs:
  - 192.168.72.100
```
apply clusterfile
```
sealos apply -f Clusterfile
```
error log
```
...
Mar 01 01:05:08 node34 kubelet[1945278]: E0301 01:05:08.956057 1945278 kubelet.go:2448] "Error getting node" err="node \"node34\" not found"
Mar 01 01:05:09 node34 kubelet[1945278]: E0301 01:05:09.057102 1945278 kubelet.go:2448] "Error getting node" err="node \"node34\" not found"
Mar 01 01:05:09 node34 kubelet[1945278]: E0301 01:05:09.157997 1945278 kubelet.go:2448] "Error getting node" err="node \"node34\" not found"
Mar 01 01:05:09 node34 kubelet[1945278]: E0301 01:05:09.228832 1945278 controller.go:144] failed to ensure lease exists, will retry in 7s, error: Get "https://apiserver.cluster.local:6443/apis/coordination.k8s.io/v1/namespaces/kube-node-lease/leases/node34?timeout=10s": x509: certificate is valid for localhost, kubernetes, kubernetes.default, kubernetes.default.svc, kubernetes.default.svc.cluster.local, node34, not apiserver.cluster.local
Mar 01 01:05:09 node34 kubelet[1945278]: E0301 01:05:09.258841 1945278 kubelet.go:2448] "Error getting node" err="node \"node34\" not found"
Mar 01 01:05:09 node34 kubelet[1945278]: I0301 01:05:09.297233 1945278 kubelet_node_status.go:70] "Attempting to register node" node="node34"
[...]

PR #2752, opened 2023-03-08
BUG: custom certSANs should append to the default ClusterConfiguration certSANs

hexinatgithub:
custom certSANs should append to the default ClusterConfiguration certSANs

### DIFF
--- a/.github/workflows/test_run.yml
+++ b/.github/workflows/test_run.yml
@@ -210,6 +210,7 @@ jobs:
           sudo cat /root/.sealos/default/Clusterfile
       - name: Verify Cluster Status
         run: |
+          local_ip=`ip addr | grep inet | grep -v inet6 | grep -v docker |grep -v host | awk '{print $2}' | awk -F '/' '{print $1}'`
           echo "Verify Cluster"
           echo "Current system info"
           sudo /var/lib/sealos/data/default/rootfs/opt/sealctl cri socket
@@ -220,6 +221,10 @@ jobs:
           sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep systemd
           sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep 100.64.0.0/10
           sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep 100.55.0.0/16
+          sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep 127.0.0.1
+          sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep apiserver.cluster.local
+          sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep 10.103.97.2
+          sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep ${local_ip}
 
   verify-run-docker-buildimage:
     needs: [build-sealos]
@@ -324,6 +329,9 @@ jobs:
           kind: ClusterConfiguration
           networking:
             serviceSubnet: "100.55.0.0/16"
+          apiServer:
+            certSANs:
+            - 192.168.72.100
           EOF
           cat /tmp/apply/Clusterfile
       - name: Auto install k8s using sealos
@@ -338,6 +346,7 @@ jobs:
           sudo cat /root/.sealos/default/Clusterfile
       - name: Verify Cluster Status
         run: |
+          local_ip=`ip addr | grep inet | grep -v inet6 | grep -v docker |grep -v host | awk '{print $2}' | awk -F '/' '{print $1}'`
           echo "Verify Cluster"
           echo "Current system info"
           sudo /var/lib/sealos/data/default/rootfs/opt/sealctl cri socket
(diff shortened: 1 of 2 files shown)


hexinatgithub:
#2688 
User may deploy an external LB to access the apiserver.

fengxsong:
I think we just need to add one more mergo configuration to solve this problem

```
var defaultMergeOpts = []func(*mergo.Config){
	mergo.WithOverride,
	mergo.WithAppendSlice,
}
```

hexinatgithub:
> I think we just need to add one more mergo configuration to solve this problem
> 
> ```
> var defaultMergeOpts = []func(*mergo.Config){
> 	mergo.WithOverride,
> 	mergo.WithAppendSlice,
> }
> ```
> 
> at the meantime, remove line https://github.com/labring/sealos/blob/main/pkg/runtime/kubeadm_config.go#L103
> 
> @cuisongliu needs some test.

And that's exactly what I thought in the beginning.
I'm not very sure other slice fields should append or overwrite it.

.github/workflows/test_run.yml
```diff
@@ -348,6 +351,7 @@ jobs:
           sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep systemd
           sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep 100.64.0.0/10
           sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep 100.55.0.0/16
+          sudo cat /root/.sealos/default/etc/kubeadm-init.yaml | grep 192.168.72.100
```
cuisongliu, line 354: need test all apiserver certSANs 

```
- 127.0.0.1
  - apiserver.cluster.local
  - 10.103.97.2
  - 10.1.0.127
  - 192.168.72.100
```

cuisongliu:
> And that's exactly what I thought in the beginning.
> I'm not very sure other slice fields should append or overwrite it.
> Maybe some slice filed need be overwrite other than use default value?

see  https://github.com/labring/sealos/actions/runs/4362210104/jobs/7626881209#step:6:585 same test ok .

.github/workflows/test_run.yml
hexinatgithub: Yes. Default value should include.

hexinatgithub:
> > And that's exactly what I thought in the beginning.
> > I'm not very sure other slice fields should append or overwrite it.
> > Maybe some slice filed need be overwrite other than use default value?
> 
> see https://github.com/labring/sealos/actions/runs/4362210104/jobs/7626881209#step:6:585 same test ok .

Don't get it. What it mean?

cuisongliu:
> Don't get it. What it mean?

https://github.com/labring/sealos/actions/runs/4362210104/jobs/7626881592#step:6:585

hexinatgithub:
> > Don't get it. What it mean?
> 
> https://github.com/labring/sealos/actions/runs/4362210104/jobs/7626881592#step:6:585

https://github.com/labring/sealos/actions/runs/4362210104/workflow#L341 Test both Clusterfile and kubeadm-init.yaml?

hexinatgithub:
> I think we just need to add one more mergo configuration to solve this problem
> 
> ```
> var defaultMergeOpts = []func(*mergo.Config){
> 	mergo.WithOverride,
> 	mergo.WithAppendSlice,
> }
> ```
> 
> at the meantime, remove line https://github.com/labring/sealos/blob/main/pkg/runtime/kubeadm_config.go#L103
> 
> @cuisongliu needs some test.

More conservative change will avoid introduce unnoticed bug.
@cuisongliu need some suggestion.

cuisongliu:
> > > Don't get it. What it mean?
> > 
> > 
> > https://github.com/labring/sealos/actions/runs/4362210104/jobs/7626881592#step:6:585
> 
> https://github.com/labring/sealos/actions/runs/4362210104/workflow#L341 Test both Clusterfile and kubeadm-init.yaml?

Mainly to test kubeadm-init.yaml, it needs to cover more comprehensively, don’t lose the original configuration.

cuisongliu:
> > I think we just need to add one more mergo configuration to solve this problem
> > ```
> > var defaultMergeOpts = []func(*mergo.Config){
> > 	mergo.WithOverride,
> > 	mergo.WithAppendSlice,
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > at the meantime, remove line https://github.com/labring/sealos/blob/main/pkg/runtime/kubeadm_config.go#L103
> > @cuisongliu needs some test.
> 
> More conservative change will avoid introduce unnoticed bug. @cuisongliu need some suggestion.

remove and test.

fengxsong:
> I'm not very sure other slice fields should append or overwrite it.

one more concern, the [`scheme.Default(obj)`](https://github.com/labring/sealos/blob/main/pkg/runtime/kubeadm_config.go#L244)  should not invoked before complete merging configs.

cuisongliu:
https://github.com/labring/sealos/blob/11c2f8fde2f6c0429771f6e4a36f1767a98d3d3c/.github/workflows/test_run.yml#L187  Also need to add test case

hexinatgithub:
Done.

cuisongliu (approved):

PR #2774, opened 2023-03-09
feat: add default time to image cr.

lingdie:


controllers/imagehub/api/v1/image_types.go
```diff
@@ -113,9 +115,10 @@ type ImageDetailInfo struct {
...
 	URL string `json:"url,omitempty"`
 	// ID Arch Size Ctime should use buildah inspect to get infor.
-	ID    string      `json:"ID,omitempty"`
-	Arch  string      `json:"arch,omitempty"`
-	Size  int64       `json:"size,omitempty"`
+	ID   string `json:"ID,omitempty"`
+	Arch string `json:"arch,omitempty"`
+	Size int64  `json:"size,omitempty"`
+	//+kubebuilder:default:="1970-01-01T00:00:00Z"
 	CTime metav1.Time `json:"creatTime"`
```
fanux, line 122: `CreateTime` is better than `CTime`.

fanux (approved):

controllers/imagehub/api/v1/image_types.go
zzjin: Or `ImageBuildTime`?

lingdie: use `CreateTime`, since json is `creatTime` and not easy to change.

zzjin (approved):

PR #2789, opened 2023-03-10
feat: add infra concurrent opts & optimize delete instances

xiaohan1202:


zzjin:
Why do we need to limit concurrent reconciles? And some context to pr message?

zzjin (approved):

PR #2791, opened 2023-03-13
feat modify request header & delete infra

zjy365:


frontend/desktop/src/services/request.ts
```diff
@@ -60,16 +60,18 @@ request.interceptors.request.use(
       config.url = process.env.NEXT_PUBLIC_SERVICE + config.url;
     }
 
-    let _headers: AxiosRequestHeaders = {};
+    let _headers: RawAxiosRequestHeaders = config.headers || {};
+    const session = useSessionStore.getState().session;
+    _headers['Authorization'] = encodeURIComponent(JSON.stringify(session?.kubeconfig) || '');
```
zzjin, line 65: session?.kubeconfig is already a yaml-encoded js string type, should we do not need to json `stringify` it?

zzjin (approved):

PR #2792, opened 2023-03-13
feature(main): add system config to sealos

cuisongliu:

### DIFF
--- a/pkg/system/config.go
+++ b/pkg/system/config.go
@@ -1,0 +1,122 @@
+/*
+Copyright 2023 cuisongliu@qq.com.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package system
+
+import (
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/containers/buildah"
+	"github.com/containers/storage/pkg/homedir"
+)
+
+type envSystemConfig struct{}
+
+func Get(key string) (string, error) {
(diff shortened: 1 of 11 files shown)


pkg/system/config.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (c *systemConfig) Get(key string) (string, error) {
+	return getValueOrDefault(key)
+}
+
+func (c *systemConfig) Set(key, value string) error {
+	return setValue(key, value)
+}
+
+func NewConfig() Config {
```
fengxsong, line 43: initialize a global config getter/setter so caller don't need to invoke function `NewConfig` repeatedly.

pkg/system/config.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/containers/buildah"
+	"github.com/containers/storage/pkg/homedir"
+)
+
+type Config interface {
+	Get(string) (string, error)
+	Set(string, string) error
+}
+
+type systemConfig struct{}
```
fengxsong, line 33: consider named it to `envSystemConfig`?

pkg/system/config_cmd.go
```diff
@@ -0,0 +1,87 @@
...
+
+package system
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func NewConfigCmd() *cobra.Command {
```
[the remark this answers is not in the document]
cuisongliu: > do we need to create a `alpha` subcommand to contains all of those experimental subcommands?

Yes, but we need to do it independently with a pr.

pkg/system/config.go
cuisongliu: done

pkg/system/config.go
```diff
@@ -0,0 +1,125 @@
...
+	RuntimeRootConfigKey   = "sealos_runtime_root"
+	DataRootConfigKey      = "sealos_data_root"
+	BuildahFormatConfigKey = "buildah_format"
+)
+
+func getValueOrDefault(key string) (string, error) {
+	for _, option := range configOptions {
+		if option.Key == key {
+			if option.OSEnv != "" {
+				if value := os.Getenv(option.OSEnv); value != "" {
```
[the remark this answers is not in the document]
cuisongliu: done

fengxsong (approved):

PR #2840, opened 2023-03-20
feat: add sudo support in ssh command

fengxsong:
In order to support running commands in remote host with a non-root user(but must be a superuser or sudoer), this PR makes changes below:
1. make properties of `Option` struct non-exportable, use `OptionFunc` instead (OPTIONAL);
2. auto input password after ask-sudo-prompt;
3. joins commands into single one command;

---

NOTE: Those commands that sealos executed in remote hosts have not been modified and need to be discussed further.


fengxsong:
for testing

```golang
package main

import (
	"os"

	"github.com/labring/sealos/pkg/ssh"
	"github.com/labring/sealos/pkg/utils/logger"
	"github.com/spf13/pflag"
)

func main() {
	fs := pflag.NewFlagSet("ssh", pflag.ExitOnError)
	async := fs.Bool("async", false, "run async")
	opt := ssh.NewOption()
	opt.BindFlags(fs)
	fs.Parse(os.Args)

	cli, err := ssh.New(opt)
	if err != nil {
		logger.Fatal(err)
	}
	if *async {
		err = cli.CmdAsync(fs.Arg(1), fs.Arg(2))
		if err != nil {
			logger.Fatal("failed to run async %s, err: %v", fs.Arg(2), err)
		}
	} else {
		out, err := cli.Cmd(fs.Arg(1), fs.Arg(2))
		if err != nil {
			logger.Fatal("failed to run %s, output: %s, err: %v", fs.Arg(2), string(out), err)
		}
		logger.Info("run", string(out))
	}

}

```

cuisongliu (approved):

PR #2841, opened 2023-03-20
add bytebase controller

dinoallo:


controllers/db/bytebase/api/v1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
...
+package v1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "db.sealos.io", Version: "v1"}
```
zzjin, line 29: Maybe using `bytebase.db.sealo.io` is better choice? since we may have multiple db apps.

controllers/db/bytebase/controllers/bytebase_controller.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (r *BytebaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := r.Logger
+	bb := &bbv1.Bytebase{}
+	// Get CRD bytebase (status). Call reconciler again if not found.
+	if err := r.Get(ctx, req.NamespacedName, bb); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if bb.ObjectMeta.DeletionTimestamp.IsZero() {
```
fanux, line 98: @cuisongliu seems we have tools to handle the finalizer.

controllers/db/bytebase/controllers/bytebase_controller.go
```diff
@@ -0,0 +1,226 @@
...
+	svc := corev1.Service{}
+	if err := r.Get(ctx, client.ObjectKey{
+		Namespace: req.Namespace,
+		Name:      bb.Name,
+	}, &svc); err != nil {
+		logger.Error(err, "get bytebase service failed")
+		return ctrl.Result{}, err
+	}
+
+	url := "http://" + svc.Spec.ClusterIP + ":" + bb.Spec.Port.String()
```
fanux, line 153: Why not use domain name?  fmt.Sprintf("http://%s:%s",,) is better.

xiao-jay:
please exec `git rebase main`to solve the ci error @dinoallo

controllers/db/bytebase/controllers/bytebase_controller.go
dinoallo: Users can change the domain name of the cluster through configuring kubelet, domain name like `*.cluster.local` is not always true for the cluster. We might need to start a conversation with kubelet to know the actual domain name(Is there a way to use kubernetes api (api-server) to achieve the same thing aside from open a proxy?). I think we should keep what api-server has to offer, i.e. the service IP.

dinoallo: > fmt.Sprintf("http://%s:%s",,) is better.

This is resolved by latest commits.

controllers/db/bytebase/controllers/sync.go
```diff
@@ -0,0 +1,288 @@
...
+			Name:      bb.Name,
+			Namespace: bb.Namespace,
+		},
+	}
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, ingress, func() error {
+		cookies, err := r.getClientCookie()
+		if err != nil {
+			return err
+		}
+		wholeSnippet := `more_clear_headers "X-Frame-Options:"; more_set_headers "Content-Security-Policy: default-src * blob: data: *.cloud.sealos.io cloud.sealos.io; img-src * data: blob: resource: *.cloud.sealos.io cloud.sealos.io; connect-src * wss: blob: resource:; style-src 'self' 'unsafe-inline' blob: *.cloud.sealos.io cloud.sealos.io resource:; script-src 'self' 'unsafe-inline' 'unsafe-eval' blob: *.cloud.sealos.io cloud.sealos.io resource: *.baidu.com *.bdstatic.com; frame-src 'self' cloud.sealos.io mailto: tel: weixin: mtt: *.baidu.com; frame-ancestors 'self' https://cloud.sealos.io https://*.cloud.sealos.io"; more_set_headers "X-Xss-Protection: 1; mode=block"; %s if ($request_uri ~* \.(js|css|gif|jpe?g|png)) { expires 30d; add_header Cache-Control "public";}`
```
zzjin, line 45: All hard coded domain `cloud.sealo.io` or `*.cloud.sealos.io` need to be formatted to real deploy env like `cloud.sealos.cn`

zzjin (changes requested):
Fix snippets

controllers/db/bytebase/controllers/sync.go
dinoallo: This has been resolved in the latest commits

zzjin (approved):

PR #2855, opened 2023-03-23
fix: ensure response body has to be read fully and close correctly; a…

### LINKED ISSUE #2784: BUG: image-shim request registry link did not break, resulting in a large number of links
### Sealos Version

v4.1.4

### How to reproduce the bug?

查看cri-shim到registry的链接数量
image
image

_No response_

### What do you see instead?

_No response_

### Operating environment

```markdown
- Sealos version:
- Docker version:
- Kubernetes version:
- Operating system:
- Runtime environment:
- Cluster size:
- Additional information:
```

### Additional information

_No response_

fengxsong:
fix #2784

### DIFF
--- a/pkg/utils/registry/registry.go
+++ b/pkg/utils/registry/registry.go
@@ -18,12 +18,33 @@ package registry
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/labring/sealos/fork/github.com/heroku/docker-registry-client/registry"
 	"github.com/labring/sealos/pkg/utils/logger"
 )
 
+type option struct {
+	url, username, password string
+	skipTLS                 bool
+}
+
+var (
+	sharedClients = make(map[option]*registry.Registry)
+	lock          sync.RWMutex
+)
+
 func NewRegistry(url, username, password string, skipTLS bool) (*registry.Registry, error) {
+	key := option{url, username, password, skipTLS}
+	lock.Lock()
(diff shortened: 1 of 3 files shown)


fengxsong:
@bxy4543 do you have time to take a look?

pkg/utils/registry/registry.go
```diff
@@ -18,12 +18,33 @@ package registry
...
+var (
+	sharedClients = make(map[*option]*registry.Registry)
+	lock          sync.RWMutex
+)
+
 func NewRegistry(url, username, password string, skipTLS bool) (*registry.Registry, error) {
+	key := &option{url, username, password, skipTLS}
+	lock.Lock()
+	defer lock.Unlock()
+	if v, ok := sharedClients[key]; ok {
```
bxy4543, line 41: This doesn't always seem ok, you should replace the pointer type with a struct type, which will compare each property value of the struct.

bxy4543:
> @bxy4543 do you have time to take a look?

This does not seem to be related to image-cri-shim and sealos.hub registry, how to solve this problem #2784

pkg/utils/registry/registry.go
fengxsong: > This doesn't always seem ok, you should replace the pointer type with a struct type, which will compare each property value of the struct.

make sense👍

bxy4543 (approved):

Issue #2832, opened 2023-03-18
Feature: run clusterfile repeatedly


willzhang:
### What is the problem this feature will solve?

run clusterfile repeatedly

sealos version: v4.1.5

### If you have solution，please describe it

basic Clusterfile
```
root@ubuntu:~# cat Clusterfile 
apiVersion: apps.sealos.io/v1beta1
kind: Cluster
metadata:
  creationTimestamp: null
  name: default
spec:
  hosts:
  - ips:
    - 192.168.72.35:22
    roles:
    - master
    - amd64
  image:
  - hub.sealos.cn/labring/kubernetes:v1.25.6
  - hub.sealos.cn/labring/helm:v3.11.0
  - hub.sealos.cn/labring/calico:v3.24.5
  ssh:
    passwd: "123456"
    pk: /root/.ssh/id_rsa
    port: 22
status: {}
```
apply
```
sealos apply -f Clusterfile 
```
add new app`hub.sealos.cn/labring/nginx:v1.23.3`  to Clusterfile 
```
apiVersion: apps.sealos.io/v1beta1
kind: Cluster
metadata:
  creationTimestamp: null
  name: default
spec:
  hosts:
  - ips:
    - 192.168.72.35:22
    roles:
    - master
    - amd64
  image:
  - hub.sealos.cn/labring/kubernetes:v1.25.6
  - hub.sealos.cn/labring/helm:v3.11.0
  - hub.sealos.cn/labring/calico:v3.24.5
  - hub.sealos.cn/labring/nginx:v1.23.3
  ssh:
    passwd: "123456"
    pk: /root/.ssh/id_rsa
    port: 22
status: {}
```
apply again, new app not installed, but i hope it installed
```
root@ubuntu:~# sealos apply -f Clusterfile 
2023-03-18T10:56:06 info no nodes that need to be scaled
2023-03-18T10:56:06 info 
      ___           ___           ___           ___       ___           ___
     /\  \         /\  \         /\  \         /\__\     /\  \         /\  \
    /::\  \       /::\  \       /::\  \       /:/  /    /::\  \       /::\  \
   /:/\ \  \     /:/\:\  \     /:/\:\  \     /:/  /    /:/\:\  \     /:/\ \  \
  _\:\~\ \  \   /::\~\:\  \   /::\~\:\  \   /:/  /    /:/  \:\  \   _\:\~\ \  \
[...]

PR #2856, opened 2023-03-24
feat: update RunNewImages based on image diffs

LZiHaN:
This commit sets the RunNewImages variable to be the difference between the current cluster's app image and the desired cluster's app image. This ensures that only the necessary changes are applied when updating the cluster. 

refer to #2832


pkg/apply/utils.go
```diff
@@ -109,3 +109,24 @@ func getHostArch(sshClient ssh.Interface) func(string) v2.Arch {
 func GetHostArch(sshClient ssh.Interface, ip string) string {
 	return string(getHostArch(sshClient)(ip))
 }
+
+func GetImagesDiff(current, desired []string) []string {
```
bxy4543, line 113: https://github.com/labring/sealos/blob/a02ebcafefbeb31b284ee8786987d7f7c6455721/pkg/utils/strings/strings.go#L174

can use this method to get different images

bxy4543 (approved):

pkg/apply/utils.go
fengxsong, line 113: IMO, we should keep the order of images, sometimes `image 1` might depends on `image 2`

cuisongliu: > IMO, we should keep the order of images, sometimes `image 1` might depends on `image 2`

Yes, it's better to keep the order while diff.

cuisongliu (changes requested):

pkg/apply/utils.go
LZiHaN: do I need to consider duplicate images in the Clusterfile?

LZiHaN: do I need to consider the uniqueness of the image for this function as it doesn't seem to have been taken into account?

pkg/apply/apply.go
```diff
@@ -56,10 +56,12 @@ func NewApplierFromFile(path string, args *Args) (applydrivers.Interface, error)
 	}
 	currentCluster := cf.GetCluster()
 
+	runNewImages := GetImagesDiff(currentCluster.Spec.Image, cluster.Spec.Image)
```
mouuii, line 59: before you call GetImagesDiff() ，please use hashfunc to quick compare the spec!

cuisongliu: > before you call GetImagesDiff() ，please use hashfunc to quick compare the spec!

agree

LZiHaN: > > before you call GetImagesDiff() ，please use hashfunc to quick compare the spec!
> 
> agree

done

pkg/apply/utils.go
LZiHaN: > IMO, we should keep the order of images, sometimes `image 1` might depends on `image 2`

done

pkg/apply/utils.go
```diff
@@ -109,3 +111,24 @@ func getHostArch(sshClient ssh.Interface) func(string) v2.Arch {
...
+	return stringsutil.RemoveDuplicate(stringsutil.RemoveStrSlice(desired, current))
+}
+
+func CompareImageSpecHash(currentImages []string, newImages []string) bool {
```
cuisongliu, line 119: hash.ToString replace it.

LZiHaN: > hash.ToString replace it.

done

cuisongliu (approved):

cuisongliu:
make format to code

cuisongliu (changes requested):
fix to pass ci

cuisongliu:
```
ok  	github.com/labring/sealos/pkg/env	0.052s	coverage: 74.5% of statements
[188](https://github.com/labring/sealos/actions/runs/4529171618/jobs/7977085728?pr=2856#step:5:189)
--- FAIL: Test_NewApplierFromFile (0.01s)
[189](https://github.com/labring/sealos/actions/runs/4529171618/jobs/7977085728?pr=2856#step:5:190)
    --- FAIL: Test_NewApplierFromFile/success (0.01s)
```
Test_NewApplierFromFile fail

LZiHaN:
> ```
> ok  	github.com/labring/sealos/pkg/env	0.052s	coverage: 74.5% of statements
> [188](https://github.com/labring/sealos/actions/runs/4529171618/jobs/7977085728?pr=2856#step:5:189)
> --- FAIL: Test_NewApplierFromFile (0.01s)
> [189](https://github.com/labring/sealos/actions/runs/4529171618/jobs/7977085728?pr=2856#step:5:190)
>     --- FAIL: Test_NewApplierFromFile/success (0.01s)
> ```
> 
> Test_NewApplierFromFile fail

Through testing, it is found that the error is caused by currentCluster.Spec. I need to modify it.

cuisongliu:
resolve your Conflicting files

LZiHaN:
> resolve your Conflicting files

done

cuisongliu (approved):

PR #2857, opened 2023-03-24
feat. registry add pull limit.

lingdie:


service/hub/utils/map.go
```diff
@@ -0,0 +1,48 @@
+package utils
+
+import (
+	"sync"
+	"time"
+)
+
+type MapWithReset struct {
+	mu    sync.Mutex
+	items map[interface{}]int64
```
cuisongliu, line 10: using any?

lingdie: ok.

service/auth/main_test.go
```diff
@@ -19,7 +19,7 @@ import (
 	"testing"
 )
 
-func TestAuthServer(t *testing.T) {
+func TestAuthServer(_ *testing.T) {
```
mouuii, line 22: Test for what ， you should enhance it not ,remove the t

service/hub/main.go
```diff
@@ -24,16 +24,22 @@ func (rs *RestartableServer) Serve(c *server.Config) {
 		Addr:    c.Server.ListenAddress,
 		Handler: as,
 	}
+
+	// start the pull request counter reset timer
+	glog.Infof("Pull request counter reset timer started")
+	go as.PullReqCounter.StartResetTimer(c.Server.PullReqCounterResetInterval)
+
```
mouuii, line 31: put the goroutine under as ,not exposed in main file

service/hub/server/config.go
```diff
@@ -83,6 +96,15 @@ func LoadConfig(fileName string) (*Config, error) {
 	if c.Server.ListenAddress == "" {
 		c.Server.ListenAddress = ":5001"
 	}
+	if c.Server.MaxRequestsPerIP == 0 {
+		c.Server.MaxRequestsPerIP = DefaultMaxRequestsPerIP
```
mouuii, line 100: can you refactor to pointer ?

service/hub/server/server.go
```diff
@@ -358,3 +369,26 @@ func (as *AuthServer) Stop() {
...
+			return true
+		}
+	}
+
+	// nomal user pull/push request
+	as.PullReqCounter.Increment(ar.RemoteIP.String())
+	as.PullReqCounter.Increment(ar.Account)
+
+	return as.PullReqCounter.Get(ar.RemoteIP.String()) < as.config.Server.MaxRequestsPerIP &&
+		as.PullReqCounter.Get(ar.Account) < as.config.Server.MaxRequestsPerAccount
```
mouuii, line 393: can you use ratelimter pkg instead of writing it ?

lingdie: ok

service/auth/main_test.go
lingdie: Not related to this pr, golint updated and I have to replace `t` to `_`, will fix this latter.

service/hub/server/config.go
lingdie: `server` refactor to a poniter?

service/hub/server/server.go
lingdie: One question about `golang.org/x/time/rate` (not sure if it is the pkg): I have to mantain a limiter for each user and ip, it's a huge cost since user and ip will increace.

service/hub/server/config.go
mouuii: if server.ListenAddress == nil like this

service/hub/server/server.go
mouuii: I think  the golang.org/x/time/rate is enough for  now , you can controller the total cost by limiter .

service/hub/utils/map.go
```diff
@@ -0,0 +1,46 @@
+package utils
+
+import (
+	"sync"
+	"time"
+)
+
+type MapWithReset struct {
```
bxy4543, line 8: This map structure seems to be designed for counting, so it should be named with related functions, like CountMap。

service/hub/main.go
lingdie: ok

service/auth/main_test.go
lingdie: seems not necessary.

service/hub/server/server.go
lingdie: ok, thx for advice.

service/hub/utils/map.go
lingdie: refactor to limiter, this map is deprecated.

service/hub/server/config.go
lingdie: not necessary？

Issue #2863, opened 2023-03-24
Feature: Refactoring Sealos to Support k3s, k0s Runtime


cubxxw:
## Problem Description

Sealos currently does not support the k3s runtime and requires the redesign of the runtime interface to avoid excessive coupling with the Processor and to add verification.

Here is a record of [testing various k3s installation methods](https://docker.nsddd.top/Cloud-Native-k8s/15.html). If you're curious about k3s, be sure to read [this note](https://docker.nsddd.top/Cloud-Native-k8s/14.html), as well as k3s • **[official](https://www.rancher.cn/k3s/) and [documentation](https://docs.rancher.cn/)**, and its GitHub • **[open source address](https://github.com/k3s-io/k3s/)**.

<br>

#### Design a new runtime interface

Redesign the runtime interface to avoid excessive coupling with the Processor and add verification.

#### Design ideas for the k3s runtime module

To support the k3s runtime, we need to add a new runtime module. This module needs to support the following functions:

+ Execute k3s installation commands
+ Execute k3s upgrade commands
+ Execute k3s uninstall commands

#### Design the k3s directory and filesystem distribution using Sealos' image distribution logic

To support the k3s runtime, we need to use Sealos' image distribution logic to design the k3s directory and filesystem distribution of the k3s image.

#### Optimize the packaging of kubeadm and k3s binaries

To support the k3s runtime, we need to optimize the packaging of kubeadm and k3s binaries.

#### Provide unit test and CI flow to ensure code quality

To ensure code quality, we need to provide unit testing and CI flow.

<br>

## Tasks

| Type          | Improvements                                                 | Issue        | PR    | Status |
| ------------- | ------------------------------------------------------------ | ------------ | ----- | ------ |
| 🚀 Feature     | Design a new runtime interface to avoid excessive coupling with the Processor and add verification | #1943, #2037 |  #3741,#3691     | ✅      |
| 🔨 Enhancement | Implement the k3s runtime module to support k3s installation, upgrade, and uninstall commands |              |       | ❌      |
| 🔧 Maintenance | Design the k3s directory and filesystem distribution using Sealos' image distribution logic |              |       | ❌      |
| 👌 Refactoring | Optimize the packaging of kubeadm and k3s binaries           |              |       | ❌      |
| 🧪 Testing     | Provide unit testing and CI flow to ensure code quality      |              |       | ❌      |
[...]

Issue #2037, opened 2022-11-04
Feature: sealos support k3s


maslow:
### What is the problem this feature will solve?

- Brings all the benefits of k3s. @https://www.rancher.cn/k3s/
- No k3s no sealos.
> Sealos wants to hit the mass market, which is consistent with the direction of k3s. The lightweight k3s is more popular with mass developers.

### If you have solution，please describe it

noop

### What alternatives have you considered?

- k0s

Issue #1943, opened 2022-10-23
Feature: support k0s, and build k0s cluster image


fanux:

fanux:
`sealos run k0s:latest --masters xxx --nodes xxxx --passwd xxxx`

cuisongliu:
@Fanux If the rootfs of this k3s/k0s are officially developed, align it with me.

fanux:
@3293172751 will try to provide this feature.

3293172751:
[k3s - github](https://github.com/k3s-io/k3s)，I wonder if I need a tutorial on [k3s.io](https://k3s.io/)，what should I pay attention to when I use it❓
Do I need to write the functions next to k8s to keep them **together,** or **separate❓** 
@fanux

cuisongliu:
> [k3s - github](https://github.com/k3s-io/k3s)，I wonder if I need a tutorial on [k3s.io](https://k3s.io/)，what should I pay attention to when I use it❓
> Do I need to write the functions next to k8s to keep them **together,** or **separate❓**
> @fanux

1. runtime interface need spilt kubeadm and k3s
2. k3s rootfs

cuisongliu:
reopen

PR #2864, opened 2023-03-24
refactor(runtime): sealos runtime module needs better design to meet …

cubxxw:
…future k3s and k0s...

This PR is still working on #1943, #2037, please note

Please give suggestions in this PR and follow up this PR

This is part of the plan: #2863


pkg/guest/guest.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/labring/sealos/fork/golang/expansion"
 	"github.com/labring/sealos/pkg/constants"
 	"github.com/labring/sealos/pkg/env"
-	"github.com/labring/sealos/pkg/runtime"
+	runtime "github.com/labring/sealos/pkg/runtime/kubernetes"
```
cuisongliu, line 25: Can you also change it here, the guest should not pay attention to the real-time details of the runtime

pkg/apply/processor/install.go
```diff
@@ -166,7 +167,7 @@ func (c *InstallProcessor) PreProcess(cluster *v2.Cluster) error {
 		cluster.SetMountImage(mount)
 		c.NewMounts = append(c.NewMounts, *mount)
 	}
-	runtime, err := runtime.NewDefaultRuntime(cluster, c.ClusterFile.GetKubeadmConfig())
+	runtime, err := kubernetes.NewDefaultRuntime(cluster, c.ClusterFile.GetKubeadmConfig())
```
cuisongliu, line 170: NewDefaultRuntime should be in `github.com/labring/sealos/pkg/runtime` .

cubxxw: yep

cmd/sealctl/cmd/token.go
```diff
@@ -30,7 +30,8 @@ func newTokenCmd() *cobra.Command {
 		Use:   "token",
 		Short: "token generator",
 		Run: func(cmd *cobra.Command, args []string) {
-			t, err := runtime.Generator()
+			//TODO: Whether a token is required to encapsulate an interface?
+			t, err := kubernetes.Generator()
```
fanux, line 34: So the caller should know runtime is kubernetes?

cmd/sealctl/cmd/token.go
```diff
@@ -30,7 +30,8 @@ func newTokenCmd() *cobra.Command {
...
 			if err != nil {
 				logger.Error("exec token error: " + err.Error())
 				os.Exit(1)
```
fanux, line 37: Use RunE to instead Exit

pkg/apply/gen.go
```diff
@@ -21,7 +21,7 @@ import (
 
 	"github.com/labring/sealos/pkg/apply/processor"
 	"github.com/labring/sealos/pkg/buildah"
-	"github.com/labring/sealos/pkg/runtime"
+	runtime "github.com/labring/sealos/pkg/runtime/kubernetes"
```
fanux, line 24: Maybe use NewRuntime(runtime type) is better than use kuberntes package directly.

pkg/apply/processor/create.go
```diff
@@ -89,7 +90,7 @@ func (c *CreateProcessor) PreProcess(cluster *v2.Cluster) error {
 	if err := MountClusterImages(cluster, c.Buildah); err != nil {
 		return err
 	}
-	runTime, err := runtime.NewDefaultRuntime(cluster, c.ClusterFile.GetKubeadmConfig())
+	runTime, err := kubernetes.NewDefaultRuntime(cluster, c.ClusterFile.GetKubeadmConfig())
```
fanux, line 93: runtime.NewRuntime(runtimetype string,....)

cuisongliu:
add `sealos.io.runtime="k8s"` to spec.

cmd/sealctl/cmd/token.go
cubxxw: I don't need to know. I'll just move the table of contents and adjust it later

pkg/apply/processor/create.go
cubxxw: NewRuntime --> Should I put it in runtime or outside?

PR #2896, opened 2023-04-01
feature: Add SSH option to the scale command(sealos add/delete).

lmsh7:
- Added SSH to `apply.ScaleArgs`, making it possible to specify password and other SSH options when adding nodes, and modified the corresponding test files.
- Fixed a documentation error that incorrectly stated "addCmd represents the delete command".
- Added a TODO at a specific location.
- Added a break statement to GetIPSByRole to prevent unnecessary comparisons.

### DIFF
--- a/pkg/apply/scale.go
+++ b/pkg/apply/scale.go
@@ -54,6 +54,23 @@ func NewScaleApplierFromArgs(scaleArgs *ScaleArgs, flag string) (applydrivers.In
 	if scaleArgs.Cluster.Nodes == "" && scaleArgs.Cluster.Masters == "" {
 		return nil, fmt.Errorf("the node or master parameter was not committed")
 	}
+	if scaleArgs.fs != nil {
+		if scaleArgs.fs.Changed("user") || cluster.Spec.SSH.User == "" {
+			cluster.Spec.SSH.User = scaleArgs.SSH.User
+		}
+		if scaleArgs.fs.Changed("pk") || cluster.Spec.SSH.Pk == "" {
+			cluster.Spec.SSH.Pk = scaleArgs.SSH.Pk
+		}
+		if scaleArgs.fs.Changed("pk-passwd") || cluster.Spec.SSH.PkPasswd == "" {
+			cluster.Spec.SSH.PkPasswd = scaleArgs.SSH.PkPassword
+		}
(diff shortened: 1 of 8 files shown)


fanux:
Great job! Please add a test report for this PR.

lmsh7:
> Great job! Please add a test report for this PR.
Test report:
1. Add node with wrong setting.
屏幕截图 2023-04-05 155248
2. Clusterfile after add failed.
屏幕截图 2023-04-05 153926
3. Add node with correct ssh setting and build connection successfully.
屏幕截图 2023-04-05 154114

pkg/apply/scale.go
```diff
@@ -54,6 +54,23 @@ func NewScaleApplierFromArgs(scaleArgs *ScaleArgs, flag string) (applydrivers.In
...
+		if scaleArgs.fs.Changed("pk-passwd") || cluster.Spec.SSH.PkPasswd == "" {
+			cluster.Spec.SSH.PkPasswd = scaleArgs.SSH.PkPassword
+		}
+		if scaleArgs.fs.Changed("port") || cluster.Spec.SSH.Port == 0 {
+			cluster.Spec.SSH.Port = scaleArgs.SSH.Port
+		}
+		if scaleArgs.fs.Changed("passwd") || cluster.Spec.SSH.Passwd == "" {
+			cluster.Spec.SSH.Passwd = scaleArgs.SSH.Password
+		}
+	}
```
lingdie, line 73: Currently, we may do not support having multiple passwords/pk in a cluster. If you replace `cluster.Spec.SSH` here, it may result in the modification of the original password and the inability of the original nodes to connect normally.

lingdie (commented):
Maybe you can add more test: create a cluster and add one node with different password and delete one node in the cluster, print the Cluster file and see what will happen.

pkg/apply/scale.go
fengxsong: > Currently, we may do not support having multiple passwords/pk in a cluster.

actually, we did. override the `.spec.hosts[*].ssh` field.

cuisongliu:
> Maybe you can add more test: create a cluster and add one node with different password and delete one node in the cluster, print the Cluster file and see what will happen.

fix this can merge it .

PR #2903, opened 2023-04-03
Add master fail after modify service subnet

Deapou:

### DIFF
--- a/pkg/runtime/kubeadm.go
+++ b/pkg/runtime/kubeadm.go
@@ -118,8 +118,8 @@ func (k *KubeadmRuntime) getCGroupDriver(node string) (string, error) {
 func (k *KubeadmRuntime) MergeKubeadmConfig() error {
 	k.ImageKubeVersion = k.getKubeVersionFromImage()
 	for _, fn := range []string{
-		"",                          // generate default kubeadm configs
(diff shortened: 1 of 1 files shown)


pkg/runtime/kubeadm.go
```diff
@@ -118,8 +118,8 @@ func (k *KubeadmRuntime) getCGroupDriver(node string) (string, error) {
 func (k *KubeadmRuntime) MergeKubeadmConfig() error {
 	k.ImageKubeVersion = k.getKubeVersionFromImage()
 	for _, fn := range []string{
-		"",                          // generate default kubeadm configs
-		k.getDefaultKubeadmConfig(), // merging from predefined path of file if file exists
+		"", // generate default kubeadm configs
+		filepath.Join(k.getContentData().EtcPath(), constants.DefaultInitKubeadmFileName), // merging from predefined path of file if file exists
```
bxy4543, line 122: no need to delete `k.getDefaultKubeadmConfig()`,  It is a configuration that is supported at build rootfs image and can be used for non-existent cases

bxy4543:
This will be modified to obtain the service subnet through kubeadm configuration to ensure data consistency, so no modification is needed here.

PR #2906, opened 2023-04-04
refactor: add back auto build and auto save abilities in `diff` command

fengxsong:
```bash
Examples:
  sealos diff labring/kubernetes:v1.25 labring/kubernetes:v1.26
  sealos diff -o table oci-archive:/path/of/older.tar oci-archive:/path/of/newer.tar
  sealos diff --patch --save -o patch.tar -t labring/kubernetes:patch-from-125-126 labring/kubernetes:v1.25 labring/kubernetes:v1.26

Flags:
      --diff-type string   type of object to diff, available options are [container, image, all] (default "image")
  -h, --help               help for diff
  -o, --output string      change the output format, available options are [json, table] (default "json")
      --patch              if enabled then create a named archive file contains changes that we interested
      --save               if enabled and --tag flag is specified, diff will save the built image into a oci-archive file with the name of --output
  -t, --tag string         tag name, diff will auto build a new image with the generated changes if flag specified and --patch=true
```

### DIFF
--- a/pkg/buildah/diff.go
+++ b/pkg/buildah/diff.go
@@ -210,7 +210,7 @@ func runDiff(c *cobra.Command, args []string, opts *diffOption) error {
 	if err != nil {
 		return err
 	}
-	changes, err := r.store.Changes(from, to)
+	changes, err := r.Store.Changes(from, to)
 	if err != nil {
 		return err
 	}
@@ -230,11 +230,11 @@ func runDiff(c *cobra.Command, args []string, opts *diffOption) error {
 			return fmt.Errorf("unknown output format %s", opts.out)
 		}
 	}
-	img, err := r.store.Image(args[1])
+	img, err := r.Store.Image(args[1])
 	if err != nil {
 		return err
 	}
-	target, err := r.store.DifferTarget(img.TopLayer)
+	target, err := r.Store.DifferTarget(img.TopLayer)
 	if err != nil {
 		return fmt.Errorf("failed to get diff target: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/buildah/diff.go
```diff
@@ -100,15 +102,16 @@ func changesToTable(out io.Writer, diffs []archive.Change) error {
...
+	save            string
 }
 
 func (o *patchOption) RegisterFlags(fs *pflag.FlagSet) {
-	fs.BoolVar(&o.createPatchFile, "create-patch", false, "If enabled then create a named archive file contains changes that we interested")
-	fs.BoolVar(&o.build, "build", false, "Auto build a new image with the generated changes")
-	fs.BoolVar(&o.save, "save", false, "Auto save the built image into a oci-archive file")
-	bailOnError(markFlagsHidden(fs, "create-patch", "build", "save"), "")
+	fs.BoolVar(&o.createPatchFile, "create-patch", false, "if enabled then create a named archive file contains changes that we interested")
+	fs.StringVar(&o.build, "build", "", "tag name, diff will auto build a new image with the generated changes if flag specified"+
```
fanux, line 111: Why tag name call it `build`? Provide an example about how to use those flags.

fengxsong: > Why tag name call it `build`? Provide an example about how to use those flags.

What would you recommend from a user experience perspective?

bxy4543:
Can you add cobra.command.example for the diff command .

fengxsong:
> Can you add cobra.command.example for the diff command .

done

cubxxw (approved):

PR #2926, opened 2023-04-10
fix: check whether the IPs in the parameter can be connected before generating Applier

mixinkexue:
#2398 ClusterFile should not save wrong IP, so check the args before applying a Applier

### DIFF
--- a/pkg/apply/run.go
+++ b/pkg/apply/run.go
@@ -122,9 +122,19 @@ func (r *ClusterArgs) runArgs(imageList []string, args *RunArgs) error {
 	clusterSSH := r.cluster.GetSSH()
 	sshClient := ssh.NewSSHClient(&clusterSSH, true)
 	if len(masters) > 0 {
+		for _, master := range masters {
+			if _, err := sshClient.Cmd(master, ""); err != nil {
+				return err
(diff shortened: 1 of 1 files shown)


pkg/apply/run.go
```diff
@@ -122,9 +122,19 @@ func (r *ClusterArgs) runArgs(imageList []string, args *RunArgs) error {
 	clusterSSH := r.cluster.GetSSH()
 	sshClient := ssh.NewSSHClient(&clusterSSH, true)
 	if len(masters) > 0 {
+		for _, master := range masters {
+			if _, err := sshClient.Cmd(master, ""); err != nil {
```
fengxsong, line 126: `sshClient.Ping` is just fine.

cuisongliu:
checker same check ping from ssh client

mixinkexue:
> checker same check ping from ssh client

If using the checker find this error, ClusterFile will record a wrong IP.This will affect subsequent commands

mixinkexue:
> > checker same check ping from ssh client
> 
> If using the checker find this error, ClusterFile will record a wrong IP.This will affect subsequent commands

so i want check it here

PR #2942, opened 2023-04-13
bugfix: the failure to add master after updating the certificate

bxy4543:
fix:
image

fix: #2398 
Get certSANS from kubeadm config after initializing cluster (join master, update cert)

Fix incorrect cluster state saving, skip precheck && preProcess incorrectly save cluster state.

### DIFF
--- a/pkg/apply/processor/error.go
+++ b/pkg/apply/processor/error.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2022.
+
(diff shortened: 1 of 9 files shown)


pkg/apply/processor/error.go
```diff
@@ -0,0 +1,31 @@
+package processor
```
cuisongliu, line 1: add lic

cuisongliu (approved):

PR #3077, opened 2023-05-12
feat: desktop pwa

zjy365:


c121914yu (approved):

frontend/desktop/public/locales/zh -CN/translation.json
zzjin, line 1: Maybe miss-spelled of file path? an empty space.

PR #3140, opened 2023-05-26
feat: add max limit in podline chart

moonrailgun:
Add max limit for cpu because cpu has limit(100%)

Its useful to found which service is busy

## Preview

before:
image

after:
image

## Test plan

Seriously, I didn't test this feature on a real local environment. Because it's too much trouble for frontend development (some style adjustments).

Because my changes are limited to a fixed file on the front end, I use the overrides function of chrome to directly modify the file that loads js, and it seems that there is no problem. If you are worried, you can test it again.

## Another problem

- I have no idea add memory limit, maybe we should append `max` into props and get external incoming? 
- Some value cannot found different, maybe we should hide/not render zero value? because its means nothing.

## Reference

https://echarts.apache.org/en/option.html#yAxis.max

### Summary
📈🎛️📊

This pull request fixes the y-axis scaling of the pod usage charts in the `frontend/providers/applaunchpad/src/components/PodLineChart/index.tsx` file. It improves the accuracy and readability of the charts for CPU, memory, and network usage.

> _The pod usage charts were a mess_
> _They showed wrong values and excess_
> _So `PodLineChart` got a fix_
> _To scale the y-axis right and quick_
> _And handle different network metrics_

### Walkthrough
*  Set the maximum values of the y-axes for the CPU, memory, and network usage charts to reflect the appropriate units and scales ([link](https://github.com/labring/sealos/pull/3140/files?diff=unified&w=0#diff-95713470394138dbc1827aff3c65db1b6e26de8b2e05dbaeccd7e707481bc700R41), [link](https://github.com/labring/sealos/pull/3140/files?diff=unified&w=0#diff-95713470394138dbc1827aff3c65db1b6e26de8b2e05dbaeccd7e707481bc700R86), [link](https://github.com/labring/sealos/pull/3140/files?diff=unified&w=0#diff-95713470394138dbc1827aff3c65db1b6e26de8b2e05dbaeccd7e707481bc700L121-R124)) in `frontend/providers/applaunchpad/src/components/PodLineChart/index.tsx`

### DIFF
--- a/frontend/providers/applaunchpad/src/components/PodLineChart/index.tsx
+++ b/frontend/providers/applaunchpad/src/components/PodLineChart/index.tsx
@@ -38,6 +38,7 @@ const PodLineChart = ({
         global: false // 缺省为 false
       },
       lineColor: '#36ADEF',
+      max: 100,
       formatter: (e: any[]) => `${((e[0].value / cpu) * 100).toFixed(2)}%`
     },
     memory: {
@@ -82,6 +83,7 @@ const PodLineChart = ({
         global: false // 缺省为 false
       },
       lineColor: '#00A9A6',
+      max: 100,
       formatter: (e: any[]) => `${((e[0].value / cpu) * 100).toFixed(2)}%`
     },
     deepBlue: {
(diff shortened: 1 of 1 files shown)


zzjin:
Thank you for contribution. Will have @c121914yu to help slove memory limit.

moonrailgun:
I think if the service is not available, the display on the graph should not be 0 instead of rendering

like this:
image

c121914yu (approved):

PR #3150, opened 2023-05-26
fix: fix loop call which occur in throw error in onmessage callback

moonrailgun:
if `source` is self, will occur loop in
windowMessage -> replyAppMessage -> (send to self) -> windowMessage

## Preview

before:

image

yes, its endless loop

after:

image

### Summary
🛡️🗣️🚫

Fix a bug in `frontend/client-sdk/src/master.ts` that caused the master SDK to reply with an error message to its own app messages. This improves the communication between the master SDK and the app SDK.

> _`source` and `window`_
> _must not be the same, or else_
> _error in autumn_

### Walkthrough
*  Add a condition to avoid sending error messages to the same window that originated the app request ([link](https://github.com/labring/sealos/pull/3150/files?diff=unified&w=0#diff-a6c87f8f20ac55bd024ec33bffb6c4c5358b48e6b3083a702281ccb19de0212dL58-R78))

### DIFF
--- a/frontend/client-sdk/src/master.ts
+++ b/frontend/client-sdk/src/master.ts
@@ -21,7 +21,8 @@ class MasterSDK {
     message = '',
     data = {}
   }: MasterReplyMessageType) {
-    if (!source) return;
+    // if not define source or source is self(Not need send message to self). Skip it
+    if (!source || source === window) return;
 
     source.postMessage(
       {


c121914yu:
@moonrailgun thank your callback.
You can put the judgment in the reply method, like this:
image

moonrailgun:
> @moonrailgun thank your callback. You can put the judgment in the reply method, like this: image

Thanks for your comment, but I'm not sure if there is a case of event delivery under the same window.

The problem I solve is to avoid the cycle of events rather than restricting the event delivery that cannot be carried out with the same window.

If your answer is yes, then I will revise my PR.

waiting for your reply.

c121914yu:
> > @moonrailgun thank your callback. You can put the judgment in the reply method, like this: image
> 
> Thanks for your comment, but I'm not sure if there is a case of event delivery under the same window.
> 
> The problem I solve is to avoid the cycle of events rather than restricting the event delivery that cannot be carried out with the same window.
> 
> If your answer is yes, then I will revise my PR.
> 
> waiting for your reply.

Putting the source determination in the reply method is fine.
All messages will be sent through this method. 
This will simplify the code.

moonrailgun:
Updated. pls review it again

@c121914yu

c121914yu (approved):

c121914yu (approved):

Issue #3402, opened 2023-06-26
Feature: sudo support for scp


cuisongliu:
### What is the problem this feature will solve?

At present, the function of sudo is not supported for the time being, because the logic of scp cannot create directories.
Fortunately, I saw some supporting related documents and records:

https://docker.nsddd.top/Cloud-Native/28.html
https://github.com/cubxxw/sealer/blob/929ea57233cb1d7dac0564acc3846f2a85fe057d/utils/ssh/connect.go#L195

After discussion, this solution is still feasible for sealos, and only needs to be encapsulated once on the scp client.

PR #3421, opened 2023-06-28
Feature:sudo support for scp

ghostloda:
Add sudo support for scp  https://github.com/labring/sealos/issues/3402.

After this, wen can use sudo user to install k8s
```
sudo sealos run labring/kubernetes:v1.26.1 labring/helm:v3.10.3 labring/calico:v3.25.0 --masters 10.29.13.21 --nodes 10.29.13.22 -u sudouser -p xxx
```

TEST scp command
image

image

### DIFF
--- a/pkg/ssh/suftp.go
+++ b/pkg/ssh/suftp.go
@@ -1,0 +1,169 @@
+/*
+Copyright 2023 ghostloda
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ssh
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/pkg/sftp"
+	"golang.org/x/crypto/ssh"
+)
+
+const (
+	promptpwd      string = "sudo password"
+	promptsuccess  string = "sudo success!"
+	prompterror    string = "sudo failed!!"
+	promptlength   int    = 13
+	sshdConfigPath        = "/etc/ssh/sshd_config"
+)
+
+func NewSudoSftpClient(conn *ssh.Client, sudopwd string, opts ...sftp.ClientOption) (*sftp.Client, error) {
+	s, err := conn.NewSession()
+	if err != nil {
+		return nil, err
+	}
+
+	sftpServerPath, err := getSftpServerPath(conn, sudopwd)
+	if err != nil {
+		return nil, err
+	}
+
+	// sudo -p specifies the password prompt from sudo. The prompt is sent to stderr
+	// sudo -S dictates that password should be read from stdin
+	// sudo -u specifies the user
+	// sh -c 'cmd' passes cmd to sh
+	// >&2 echo "%s" echos promptsucess to stderr before starting sftp server ( see https://stackoverflow.com/a/23550347 )
+	// & %s is the final command to be executed is sudo is successful, which is the path to the sftp-server binary. It is a list of possible binary paths and it will pick the first one that exists. At the end, it will look in $PATH.
+	// || sh -c '>&2 echo "%s" & exit 1' -> runs if sudo failed and ensures that prompterror is written to stderr before erroring
+	// In total this means that we can read from stderr and write to stdin to
+	cmd := fmt.Sprintf(
+		`sudo -p '%s' -S sh -c '>&2 echo "%s" & %s' || sh -c '>&2 echo "%s" & exit 1'`,
+		promptpwd+"\n",
+		promptsuccess,
+		sftpServerPath,
+		prompterror,
+	)
+
+	stdin, err := s.StdinPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	stdout, err := s.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	stderr, err := s.StderrPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	err = s.Start(cmd)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// Sudo might output a lecture, so looping for either password, error or success prompt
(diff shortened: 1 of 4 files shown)


pkg/ssh/suftp.go
```diff
@@ -0,0 +1,145 @@
...
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/pkg/sftp"
+	"golang.org/x/crypto/ssh"
+)
+
+const promptpwd string = "gimmeyourpwdnow"
```
fanux, line 29: “give me your password now” ，Add spaces for better readability.

pkg/ssh/suftp.go
```diff
@@ -0,0 +1,145 @@
...
+
+func NewSudoSftpClient(conn *ssh.Client, sudopwd string, opts ...sftp.ClientOption) (*sftp.Client, error) {
+	s, err := conn.NewSession()
+	if err != nil {
+		return nil, err
+	}
+
+	// serverpaths are the most likely paths to the sftp-server binary, ordered from most likely less likely
+	// paths are from https://winscp.net/eng/docs/faq_su#fn2
+	serverpaths := []string{
```
fanux, line 42: If the FTP server is not at this path, will there be compatibility issues?

fanux (approved):

pkg/ssh/suftp.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/pkg/sftp"
+	"golang.org/x/crypto/ssh"
+)
+
+const promptpwd string = "gimmeyourpwdnow"
+const promptsuccess string = "thesudopwdwasok"
+const prompterror string = "ohmgsudofailed!"
+const promptlength int = 15
+
+func NewSudoSftpClient(conn *ssh.Client, sudopwd string, opts ...sftp.ClientOption) (*sftp.Client, error) {
```
cuisongliu, line 34: check serverpaths is exsits?

cuisongliu (changes requested):

pkg/ssh/suftp.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// sudo -p specifies the password prompt from sudo. The prompt is sent to stderr
+	// sudo -S dictates that password should be read from stdin
+	// sudo -u specifies the user
+	// sh -c 'cmd' passes cmd to sh
+	// >&2 echo "%s" echos promptsucess to stderr before starting sftp server ( see https://stackoverflow.com/a/23550347 )
+	// & %s is the final command to be executed is sudo is successful, which is the path to the sftp-server binary. It is a list of possible binary paths and it will pick the first one that exists. At the end, it will look in $PATH.
+	// || sh -c '>&2 echo "%s" & exit 1' -> runs if sudo failed and ensures that prompterror is written to stderr before erroring
+	// In total this means that we can read from stderr and write to stdin to
+	cmd := fmt.Sprintf(
```
fengxsong, line 59: Using this approach may be enough https://github.com/sealerio/sealer/blob/main/utils/ssh/connect.go#L195

btw, do we really need to support ask-pass scenario?

pkg/ssh/suftp.go
ghostloda: replaced with **grep serverpaths  from '/etc/ssh/sshd_config'** , if file not found return error.

pkg/ssh/suftp.go
ghostloda: Most of the production environments do not grant sudo user NOPASSWD privileges.

pkg/ssh/suftp.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func getSftpServerPath(conn *ssh.Client, sudopwd string) (string, error) {
+	ses, err := conn.NewSession()
+	if err != nil {
+		return "", err
+	}
+	defer ses.Close()
+
+	cmd := fmt.Sprintf(`sudo -S grep -oP "Subsystem\s+sftp\s+\K.*" %s`, sshdConfigPath)
```
fengxsong, line 151: sudo is not required for this command

pkg/ssh/suftp.go
```diff
@@ -0,0 +1,169 @@
...
+	sshdConfigPath        = "/etc/ssh/sshd_config"
+)
+
+func NewSudoSftpClient(conn *ssh.Client, sudopwd string, opts ...sftp.ClientOption) (*sftp.Client, error) {
+	s, err := conn.NewSession()
+	if err != nil {
+		return nil, err
+	}
+
+	sftpServerPath, err := getSftpServerPath(conn, sudopwd)
```
fengxsong, line 43: and perhaps here should use a cache, we don't want to get sftp-server path from remote ssh client each time when creating sftp session.

pkg/ssh/suftp.go
ghostloda: > sudo is not required for this command

```
[root@node1 ~]# ll /etc/ssh/sshd_config 
-rw-------. 1 root root 3905 Aug  3  2022 /etc/ssh/sshd_config
```

pkg/ssh/suftp.go
ghostloda: > and perhaps here should use a cache, we don't want to get sftp-server path from remote ssh client each time when creating sftp session.

Is sftp-client caching better?

pkg/ssh/suftp.go
fengxsong: weird. the permission of this file on most of our servers are 644.

pkg/ssh/suftp.go
fengxsong: of course yes, but most of the methods needs refactored.

ghostloda: Is it enough to add a cache under this function?

```
func (c *Client) newClientAndSftpClient(host string) (*ssh.Client, *sftp.Client, error) {
	var (
		sshClient  *ssh.Client
		sftpClient *sftp.Client
		err        error
	)
	sshClient, err = c.connect(host)
	if err != nil {
		return nil, nil, err
	}
	// create sftp client
	if c.Option.sudo || c.Option.user != defaultUsername {
		sftpClient, err = NewSudoSftpClient(sshClient, c.password)
	} else {
		sftpClient, err = sftp.NewClient(sshClient)
	}

	return sshClient, sftpClient, err
}
```

ghostloda: How about add a new pr to add a client cache? Like this approach https://github.com/sealerio/sealer/blob/main/utils/ssh/connect.go#L142

fanux (approved):

cuisongliu (approved):

Issue #3445, opened 2023-06-30
upgrade k8s failed on old versions


yangxggo:
### Sealos Version

v4.2.3

### How to reproduce the bug?

1. deploy kubernetes or kubernetes-docker by sealos( before or include k8s v1.23.x version)
eg: 
`sealos run labring/kubernetes:v1.23.17-4.2.3 labring/helm:v3.11.0 labring/calico:v3.24.1 --masters xxxx --nodes yyyy -p zzzz`
3. upgrade k8s version to v1.24
`sealos run labring/kubernetes:v1.24.15-4.2.3`

### What is the expected behavior?

upgrade k8s version to v1.24.15

### What do you see instead?

failed with log:

Error: configmaps "kubelet-config" not found

### Operating environment

```markdown
- Sealos version:
- Docker version:
- Kubernetes version:
- Operating system: 
- Runtime environment:
- Cluster size:
- Additional information:
```

yangxggo:
k8s uses dynamic name for kubelet's configmap before release1.23(include) and it uses fixed name from release 1.24

PR #3447, opened 2023-06-30
support old version of kubelet-config

yangxggo:
associate issue #3445 

sealos checks kubelet configmap when it runing upgrade k8s verison.

### DIFF
--- a/pkg/client-go/kubernetes/expansion.go
+++ b/pkg/client-go/kubernetes/expansion.go
@@ -19,14 +19,16 @@ package kubernetes
 import (
 	"context"
 	"fmt"
-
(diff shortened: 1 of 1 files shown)


yangxggo:
k8s uses dynamic name for kubelet's configmap before release1.23(include) and it uses fixed name from release 1.24.

my solution:
1. check "kubelet-config" configMap as before
2. if not found, check configmap with prefix "kubelet-config-"
3. if found, get the data from it and clone the data to a new configmap named "kubelet-config" 

after that, it will upgrade successfully and work well in the future upgrades

cuisongliu (approved):

PR #3450, opened 2023-06-30
feat: add host ssh and sftp client map cache for sftp

ArenaSu:
Add host ssh and sftp client map cache for sftp action. This feature is disscussed here [sudo support for scp](https://github.com/labring/sealos/pull/3421#discussion_r1246021336)


cuisongliu (approved):

ArenaSu:
> need test many node join.

Test done, a cluster which had 3 nodes was installed successfully via sealos.

PR #3501, opened 2023-07-13
docs: add install example for alist && discuz && pingvin-share

nightwhite:
install example for alist && discuz && pingvin-share


bxy4543 (approved):

fanux:
Seems can't rerun the failed job.  Put them into application store seems a better way.

Issue #3428, opened 2023-06-28
BUG: sealos run status error when once running 


cuisongliu:
### Sealos Version

v4.2.2

### How to reproduce the bug?

sealos run kubernetes.tar helm.tar calico.tar cert-manager.tar zot.tar kubeblocks.tar 

when kb running error, the cluster status is failed

<img width="382" alt="img_v2_d2659541-59de-496f-a94c-40e7ae73d27g" src="https://github.com/labring/sealos/assets/4150905/51681acc-b28f-4708-a0bd-2aef66a5f05f">

### Operating environment

```markdown
- Sealos version:
- Docker version:
- Kubernetes version:
- Operating system:
- Runtime environment:
- Cluster size:
- Additional information:
```

PR #3572, opened 2023-07-24
fix cluster status

ghostloda:
fix issue https://github.com/labring/sealos/issues/3428

### DIFF
--- a/pkg/apply/processor/errors.go
+++ b/pkg/apply/processor/errors.go
@@ -16,6 +16,20 @@ limitations under the License.
 
 package processor
 
-import "errors"
+import (
+	"errors"
+	"strings"
+)
+
+const (
+	RunGuestFailed = "RunGuestFailed"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/apply/processor/errors.go
```diff
@@ -16,6 +16,20 @@ limitations under the License.
...
+
+const (
+	RunGuestFailed = "RunGuestFailed"
+)
+
+func IsRunGuestFailed(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), RunGuestFailed)
```
cuisongliu, line 32: Is it better to adjust strings.Contains to strings.HasPrefix ？

ghostloda: > Is it better to adjust strings.Contains to strings.HasPrefix ？
Considering that this error is wraped elsewhere, the Contains is more general.

cuisongliu (approved):

PR #3754, opened 2023-08-23
Count pods that not in succeeded & watch gpu changes on nodes

nowinkeyy:
_Null_

### DIFF
--- a/controllers/node/controllers/gpu_controller.go
+++ b/controllers/node/controllers/gpu_controller.go
@@ -253,7 +253,7 @@ func (r *GpuReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		if _, ok := pod.Spec.NodeSelector[NvidiaGPUProduct]; !ok {
 			return nil
 		}
-		if pod.Status.Phase != corev1.PodRunning {
+		if pod.Status.Phase == corev1.PodSucceeded {
 			return nil
 		}
 		return []string{GPU}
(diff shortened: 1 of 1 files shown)


controllers/node/controllers/gpu_controller.go
```diff
@@ -277,5 +281,25 @@ func (r *GpuReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				return ok
 			},
 		})).
+		Watches(&source.Kind{Type: &corev1.Node{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(predicate.Funcs{
+			CreateFunc: func(event event.CreateEvent) bool {
```
bxy4543, line 285: xxxFunc := func(event)bool {
 _, ok1 := event.Object.(*corev1.Node).Labels[NvidiaGPUMemory]
				_, ok2 := event.Object.(*corev1.Node).Labels[NvidiaGPUProduct]
				_, ok3 := event.Object.(*corev1.Node).Status.Allocatable[NvidiaGPU]
}

CreateFunc: xxxFunc
DeleteFunc: xxxFunc

same func, not repeatly

nowinkeyy: Because the parameters of Create Func and Delete Func are different, it is not easy to abstract into a method?

bxy4543: > Because the parameters of Create Func and Delete Func are different, it is not easy to abstract into a method?

With the same implementation, it can definitely be extracted：
```
xxxFunc := func(obj Object)bool {
_, ok1 := obj.(*corev1.Node).Labels[NvidiaGPUMemory]
_, ok2 := obj.(*corev1.Node).Labels[NvidiaGPUProduct]
_, ok3 := obj.(*corev1.Node).Status.Allocatable[NvidiaGPU]
}
CreateFunc: func(event xxEvent) bool { xxxFunc(event)}
DeleteFunc: func(event xxEvent) bool { xxxFunc(event)}
```

bxy4543 (approved):

PR #3903, opened 2023-09-12
feature: sync .sealos workdir

ghostloda:
support https://github.com/labring/sealos/issues/3859

### DIFF
--- a/pkg/apply/applydrivers/apply_drivers_default.go
+++ b/pkg/apply/applydrivers/apply_drivers_default.go
@@ -27,6 +27,7 @@ import (
 	"github.com/labring/sealos/pkg/client-go/kubernetes"
 	"github.com/labring/sealos/pkg/clusterfile"
 	"github.com/labring/sealos/pkg/constants"
+	"github.com/labring/sealos/pkg/ssh"
 	v2 "github.com/labring/sealos/pkg/types/v1beta1"
 	"github.com/labring/sealos/pkg/utils/confirm"
 	"github.com/labring/sealos/pkg/utils/iputils"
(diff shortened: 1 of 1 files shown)


pkg/apply/applydrivers/apply_drivers_default.go
```diff
@@ -273,3 +268,30 @@ func (c *Applier) deleteCluster() error {
 	logger.Info("succeeded in deleting current cluster")
 	return nil
 }
+
+func (c *Applier) syncWorkdir() {
+	workDir := constants.ClusterDir(c.ClusterDesired.Name)
+	sshCmd, err := ssh.NewExecCmdFromRoles(c.ClusterDesired, v2.MASTER)
```
cuisongliu, line 274: if master number is one, same not sync it ?

pkg/apply/applydrivers/apply_drivers_default.go
```diff
@@ -273,3 +268,30 @@ func (c *Applier) deleteCluster() error {
 	logger.Info("succeeded in deleting current cluster")
 	return nil
 }
+
+func (c *Applier) syncWorkdir() {
+	workDir := constants.ClusterDir(c.ClusterDesired.Name)
```
cuisongliu, line 273: if other node is not same user name ?

pkg/apply/applydrivers/apply_drivers_default.go
ghostloda: it will skip when local cp.

pkg/apply/applydrivers/apply_drivers_default.go
ghostloda: yes !,it should be ~/. sealos/.

ghostloda: > if exec sealos user is root and other node user name is not root , same this dir not using all nodes.
I'll make some modifications to this code

cuisongliu (approved):

PR #3991, opened 2023-09-25
feature(main): reset cluster not delete nodes

cuisongliu:
### Summary
🐛🛠️🌱

Fix a bug in `pkg/runtime/k3s/lifecycle.go` that could remove a non-master node from the masterIPs slice. Enhance the k3s lifecycle management.

> _`masterIPs` slice_
> _bug fixed with a condition_
> _autumn leaves cluster_

### Walkthrough
* Fix a bug where a non-master node could be removed from the masterIPs slice when deleting a node ([link](https://github.com/labring/sealos/pull/3991/files?diff=unified&w=0#diff-18b1fd2bd860036843fa344a0cd7d1ab962be73078b9afdad30e9835009da6afL66-R69) in `pkg/runtime/k3s/lifecycle.go`)


cuisongliu:
k3s reset not care k3s node resource.

PR #4161, opened 2023-10-24
refactor(main): unmount override container

cuisongliu:
copilot:all


pkg/apply/processor/install.go
```diff
@@ -140,17 +140,46 @@ func (c *InstallProcessor) PreProcess(cluster *v2.Cluster) error {
 			imageTypes.Insert(string(v2.AppImage))
 		}
 	}
+	syncMounts := make([]v2.MountImage, 0)
+	imageKeys := sets.NewString()
+	if cluster.Status.Mounts != nil {
+		for _, mot := range cluster.Status.Mounts {
+			if imageKeys.Has(mot.ImageName) {
```
fengxsong, line 147: should retrieve the latest one?

pkg/types/v1beta1/utils.go
```diff
@@ -117,29 +117,13 @@ func (c *Cluster) GetRootfsImage() *MountImage {
...
-		for i, img := range c.Status.Mounts {
-			if img.Name == mount.Name && img.Type == mount.Type {
-				c.Status.Mounts[i] = *mount.DeepCopy()
-				return
-			}
+			return i, &img
 		}
-		c.Status.Mounts = append(c.Status.Mounts, *mount)
 	}
+	return 0, nil
```
fengxsong, line 126: if not found, then return -1, nil

pkg/apply/processor/install.go
```diff
@@ -168,8 +197,11 @@ func (c *InstallProcessor) PreProcess(cluster *v2.Cluster) error {
 			return err
 		}
 		mount.Env = maps.Merge(mount.Env, c.ExtraEnvs)
-
-		cluster.SetMountImage(mount)
+		if imageKeys.Has(mount.ImageName) {
```
fengxsong, line 200: if index >= 0

pkg/apply/processor/install.go
```diff
@@ -140,17 +140,46 @@ func (c *InstallProcessor) PreProcess(cluster *v2.Cluster) error {
...
+		index, mount := cluster.FindImage(img)
 		var ctrName string
-		mount := cluster.FindImage(img)
 		if mount != nil {
 			if !ForceOverride {
 				continue
 			}
-			logger.Debug("trying to override app %s", img)
-			if err := c.Buildah.Delete(mount.Name); err != nil {
+			allContainers, err := c.Buildah.ListContainers()
```
fengxsong, line 162: why do we need to remove all the containers that referenced to this image?

cuisongliu: If there are a large number of registry directories, they will occupy disk space. For example, if the image of kb is 4GB, it will be full if it is executed several times.

fengxsong: Would executing `sealos rm/rmi` commands periodically be more appropriate? We should keep as many run records as possible instead of cleaning it up automatically.

cuisongliu: The contents of all containers are basically the same. There is no need to keep it. Just keep the latest one.

pkg/apply/processor/install.go
```diff
@@ -168,8 +181,11 @@ func (c *InstallProcessor) PreProcess(cluster *v2.Cluster) error {
 			return err
 		}
 		mount.Env = maps.Merge(mount.Env, c.ExtraEnvs)
-
-		cluster.SetMountImage(mount)
+		if index >= 0 {
+			cluster.Status.Mounts[index] = *mount
+			continue
+		}
+		cluster.Status.Mounts = append(cluster.Status.Mounts, *mount)
```
fengxsong, line 188:
```golang
if index >= 0 {
        cluster.Status.Mounts = append(cluster.Status.Mounts[:index], cluster.Status.Mounts[index+1:]...)
}
cluster.Status.Mounts = append(cluster.Status.Mounts, *mount)
```

pkg/apply/processor/install.go
```diff
@@ -140,9 +140,22 @@ func (c *InstallProcessor) PreProcess(cluster *v2.Cluster) error {
...
+		if idx, exists := imageIndices[mot.ImageName]; exists {
+			// 如果ImageName已经存在，替换旧的mot
+			syncMounts[idx] = mot
+		} else {
+			// 如果ImageName不存在，添加新的mot并更新索引
+			imageIndices[mot.ImageName] = len(syncMounts)
+			syncMounts = append(syncMounts, mot)
+		}
+	}
+	cluster.Status.Mounts = syncMounts
```
fengxsong, line 155:
```golang
        // de-duplicate, append the latest record to the tail.
	mounts := make([]v2.MountImage, 0)
	indexes := make(map[string]int)
	for _, m := range cluster.Status.Mounts {
		if idx, exists := indexes[m.ImageName]; exists {
			mounts = append(mounts[:idx], mounts[idx+1:]...)
		}
		mounts = append(mounts, *m.DeepCopy())
		indexes[m.ImageName] = len(mounts) - 1
	}
	cluster.Status.Mounts = mounts
```

fengxsong: also move this code block to L189

fengxsong (changes requested):

fengxsong (approved):

PR #4342, opened 2023-11-24
fix sync error

nowinkeyy:


controllers/objectstorage/config/default/manager_auth_proxy_patch.yaml
```diff
@@ -71,7 +71,7 @@ spec:
             - name: OSInternalEndpoint
               value: "object-storage.objectstorage-system.svc.cluster.local"
             - name: OSExternalEndpoint
-              value: "objectstorage.dev.sealos.top"
+              value: "objectstorageapi.dev.sealos.top"
```
bxy4543, line 74: user tmpl env :  {{ . OSExternalEndpoint }}

PR #4448, opened 2023-12-26
Update terminal ingress

wallyxjh:


controllers/terminal/controllers/ingress.go
```diff
@@ -21,8 +21,11 @@ import (
 
 	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
```
zzjin, line 24: `pointer` is deprecated and why import to same package?

PR #4494, opened 2024-01-18
style(kubepanel): improve the overall style

mlhiter:


frontend/plugins/kubepanel/src/pages/api/kubernetes/template.ts
```diff
@@ -1,4 +1,4 @@
-import fs from 'fs';
+import fs from 'node:fs';
```
c121914yu, line 1: Why add a separate "node:"

mlhiter: solved

c121914yu (approved):

c121914yu (approved):

PR #4642, opened 2024-03-28
chore: fix mongo uri gen.

nightwhite:


lingdie:
we need 'sealos.svc' to connect mongo svc in other ns.

image

PR #4749, opened 2024-05-22
fix control-plane component status

ghostloda:
fix control-plane component status checker.

```
# sealos status
....
Error: failed to run checker: pods "kube-apiserver-test2.novalocal" not found
```

### DIFF
--- a/pkg/checker/cluster_checker.go
+++ b/pkg/checker/cluster_checker.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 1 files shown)


pkg/checker/cluster_checker.go
```diff
@@ -96,6 +99,13 @@ func (n *ClusterChecker) Check(cluster *v2.Cluster, phase string) error {
 	return n.Output(NodeList)
 }
 
+func isControlPlaneNode(node corev1.Node) bool {
+	if _, ok := node.Labels["node-role.kubernetes.io/control-plane"]; ok {
```
cuisongliu, line 103: Issues with older versions need to be considered

ghostloda: Agreed,I'll add an old version label to check.

cuisongliu (approved):

PR #4765, opened 2024-05-30
chore: add skip webhook annotation key.

lingdie:


controllers/user/api/v1/operationrequest_webhook.go
```diff
@@ -71,6 +71,10 @@ func (r ReqValidator) ValidateCreate(ctx context.Context, obj runtime.Object) er
 		return errors.New("obj convert Operationrequest is error")
 	}
 
+	if req.Annotations[OperationRequestWebhookSkipKey] == "true" {
```
bxy4543, line 74: Can annotation be null?

zzjin (approved):

PR #4821, opened 2024-06-28
Feat: impl cronjob envs

zijiren233:
image
image


zjy365:
Very good, if you can display the env, the effect will be even better.
image

zzjin (approved):

PR #4841, opened 2024-07-03
feat: add probe settings for app

zijiren233:
This PR adds the ability to configure liveness, readiness, and startup probes for apps.

**Changes:**

- Added new fields to the app edit form for configuring probes.
- Added validation for probe settings.
- Updated the app detail page to display probe information.
- Added support for probes in the `adaptYamlToEdit` and `adaptEditAppData` functions.
- Added new translations for probe-related strings.
- Added new unit tests for probe functionality.

**Screenshots:**
image
image
image
image


zzjin:
Excelent pr！ But we decide to keep launchpad simple and make users who use it are not so familar with k8s， witch means probe is an advanced acrion.

As it is also useful for those professional users，move this feature to kubepanel will be better.

PR #5360, opened 2025-02-07
feat:vlogs server add pod list.

bearslyricattack:

### DIFF
--- a/service/vlogs/server/server.go
+++ b/service/vlogs/server/server.go
@@ -1,17 +1,18 @@
 package server
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
-
-	"github.com/labring/sealos/service/pkg/api"
-	"github.com/labring/sealos/service/pkg/auth"
-	"github.com/labring/sealos/service/vlogs/request"
-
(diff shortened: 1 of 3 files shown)


service/vlogs/server/server.go
```diff
@@ -41,24 +42,91 @@ func (vl *VLogsServer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 		return
 	}
+	if req.URL.Path == "/queryPodList" {
+		err := vl.queryPodList(rw, req)
+		if err != nil {
+			http.Error(rw, fmt.Sprintf("query pod list error: %s", err), http.StatusInternalServerError)
+			log.Printf("query pod list error: %s", err)
+		}
+		return
+	}
```
bxy4543, line 52: Duplicate code, consider reusing the same function implementation:
queryConvert func( name string) return query(rw, req)

if err := queryConvert(req.URL.Path); err !=nil { return xxx}

For error output, please introduce other log component libraries, similar to the output of `logger.Error()`

bxy4543 (approved):

PR #5425, opened 2025-02-27
mv lifecycle to stand-alone directory

ghostloda:
mv lifecycle to stand-alone directory


zzjin:
You also need to adjust root dir's `go.work`(may be `go.mod`,`go.sum`is need move to lifecircle too,) to match new structure.

fanux:
It can be placed in a separate directory and maintained independently. Just ensure version compatibility in subsequent iterations. Higher versions of Sealos should be compatible with the installation of the latest five major versions of Kubernetes.

zzjin (approved):
LGTM

PR #5494, opened 2025-03-20
fix(devbox): ssh connnect info adjust

mlhiter:


mlhiter:
@ShortCipher5  Ask CR,Check my i18n English.

zzjin (approved):

Issue #5516, opened 2025-04-03
Tracking issue for Github Actions workflow files reworking


dinoallo:
According to [this post](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/), the workflow files are susceptible to leaking sensitive data like secrets of repositories.
As a result, we should make the following changes to the workflow files:
1. Use [GITHUB_TOKEN](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication) everywhere.
2. The actions should be triggered by `pull_request` or similarly less permissive event type.

This is the issue for tracking the above change. All pull requests should reference this issue.

dinoallo:
https://github.com/labring/sealos/pull/5521

dinoallo:
https://github.com/labring/sealos/pull/5585

PR #5517, opened 2025-04-07
Try fix check-linter.

zzjin:


dinoallo:
https://github.com/labring/sealos/issues/5516

Issue #5577, opened 2025-05-04
BUG: image pull failed


monshunter:
### Sealos Version

5.0.1

### How to reproduce the bug?

1. In this env:
```
Linux k8s-master-0 6.8.0-58-generic #60-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar 14 18:09:50 UTC 2025 aarch64 aarch64 aarch64 GNU/Linux
SealosVersion:
  buildDate: "2024-10-09T02:18:27Z"
  compiler: gc
  gitCommit: 2b74a1281
  gitVersion: 5.0.1
  goVersion: go1.20.14
  platform: linux/arm64
```
2. With this config (cluster.yaml):
```
apiVersion: apps.sealos.io/v1beta1
kind: Cluster
metadata:
  name: default
spec:
  hosts:
    - ips:
      - "192.168.64.15"
      roles:
        - master
        - arm64
    - ips:
      - "192.168.64.21"
      - "192.168.64.22"
      - "192.168.64.23"
      roles:
        - node
        - arm64 
  image:
    - registry.cn-shanghai.aliyuncs.com/labring/kubernetes:v1.31.0
    - registry.cn-shanghai.aliyuncs.com/labring/helm:v3.17.1
    - registry.cn-shanghai.aliyuncs.com/labring/cilium:1.16.3
  ssh:
    passwd: xxxx
    pk: /root/.ssh/id_rsa
    port: 22
    user: root
```
3. Run:
```
sealos apply -f cluster.yaml
```
4. See Error:
```
2025-05-05T00:06:40 info start to copy kubeconfig files to masters
2025-05-05T00:06:40 info start to copy static files to masters
2025-05-05T00:06:40 info start to init master0...
failed to create new CRI runtime service: validate service connection: validate CRI v1 runtime API for endpoint "unix:///var/run/image-cri-shim.sock": rpc error: code = Unimplemented desc = unknown service runtime.v1.RuntimeService
To see the stack trace of this error execute with --v=5 or higher
2025-05-05T00:06:40 error Applied to cluster error: failed to init masters: master pull image failed, error: exit status 1
Error: failed to init masters: master pull image failed, error: exit status 1
```

### Operating environment

```markdown
- Sealos version:
- Docker version:
- Kubernetes version:
- Operating system:
- Runtime environment:
- Cluster size:
- Additional information:
```

ermazi:
i dont know why  kube image pull need to check cri runtimeService interface.
[...]

PR #5580, opened 2025-05-07
修复 *v1RuntimeService does not implement 问题

ermazi:
try to fix https://github.com/labring/sealos/issues/5577


ermazi:
kubeadmin no need to use image cri shim, close it

PR #5585, opened 2025-05-12
ci: rewrite workflows

dinoallo:
1. remove stale workflow files
2. remove all robot account usage
3. avoid always building all controllers, frontends


dinoallo:
https://github.com/labring/sealos/issues/5516

PR #5727, opened 2025-07-28
container commit 

Zllinc:


Panlq:
I have a question to consult: If delete the pod first and then create a new container to perform a commit, it seems that this does not save the packages previously installed in the pod or other items that require devbox persistence. right??

PR #5814, opened 2025-08-11
feat(dbprovider)：add remark function & fix database creation time

PeiranXu-108:
1. Added remark function for the name of the database and insert the remark in the cluster.
2. Fix the database creation time: from standard UTC to Beijing Time(UTC+8).


frontend/providers/dbprovider/src/constants/db.ts
```diff
@@ -38,6 +38,8 @@ export enum DBTypeEnum {
   clickhouse = 'clickhouse'
 }
 
+export const DB_REMARK_KEY = 'cluster.sealos.io/remark';
```
zjy365: 'cloud.sealos.io/remark';

PR #5909, opened 2025-08-30
feat(apply): conditionally append components based on distribution type

cuisongliu:
This pull request introduces a small update to the cluster application driver logic, specifically targeting the handling of runtime components based on distribution type. The change ensures that components are only written back when the distribution is `k3s`, improving correctness for multi-distribution support.

Distribution-specific handling:

* Added import of the `k3s` distribution identifier from `github.com/labring/sealos/pkg/runtime/k3s` to enable distribution checks.
* Updated the `getWriteBackObjects` method in `apply_drivers_default.go` to append runtime components only if the cluster distribution is `k3s`, preventing unintended writes for other distributions.

### DIFF
--- a/lifecycle/pkg/apply/applydrivers/apply_drivers_default.go
+++ b/lifecycle/pkg/apply/applydrivers/apply_drivers_default.go
@@ -21,6 +21,8 @@ import (
 	"os"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


cuisongliu:
494cdaaa98dc73052ce49796aa44df84
need fix it.

bxy4543 (approved):

PR #5981, opened 2025-09-15
fix(checker): update time synchronization check and add CockroachDB max offset configuration

cuisongliu:


deploy/cloud/charts/database/templates/cockrachdb.yaml
```diff
@@ -4,6 +4,8 @@ metadata:
   name: sealos-cockroachdb
   namespace: sealos
 spec:
+  additionalArgs:
+  - --max-offset=2s
```
dinoallo: Is 500ms default offset too low?

dinoallo (approved):

PR #5994, opened 2025-09-18
fix(webhooks): add permissions for contents and packages in GitHub Actions workflow

cuisongliu:


.github/workflows/webhooks.yml
```diff
@@ -66,6 +66,9 @@ jobs:
 
   image-build:
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      packages: write
```
dinoallo, line 71: I am not sure if we should grant `packages: write` when  `pull_request` may be triggered

PR #5995, opened 2025-09-18
feat(desktop): show captcha on signin page

Nixieboluo:


frontend/desktop/src/types/system.ts
```diff
@@ -221,7 +218,7 @@ export type AuthClientConfigType = {
       'captcha.ali.accessKeySecret',
       'captcha.ali.endpoint',
       // turnstile
-      'turnstile.cloudflare.secretKey'
```
zjy365, line 225: Could you double-check this now that we've changed the type?

zjy365: And here's the configmap.yaml.tmpl file.

zjy365 (approved):

PR #6039, opened 2025-09-24
feat(webhook): add pod resource request auto-adjustment webhook for optimizing cluster utilization

zijiren233:
## Description

This PR introduces a new admission webhook for Pod resources that implements intelligent resource request auto-adjustment to optimize cluster resource utilization. The webhook automatically adjusts pod resource requests based on configurable ratios while preventing invalid resource configurations.

### 🚀 Resource Request Auto-Adjustment
- **Automatic Request Optimization**: Automatically adjusts pod resource requests based on limits and configurable ratios to improve cluster resource utilization
- **Differentiated Strategies**: 
  - Default pods: 1:10 limit-to-request ratio (customizable)
  - Database pods (KubeBlocks managed): 1:5 limit-to-request ratio (customizable)
- **Smart Thresholds**: Skip adjustment for pods with limits below configurable thresholds (default: 100m CPU, 128Mi memory)

### 🛡️ Resource Validation
- Prevents pods from setting CPU or memory limits to "0"
- Validates both regular containers and init containers
- Special handling for database pods (only validates first container)

### 🎯 Targeted Application
- Only applies to user namespaces (namespaces starting with "ns-")
- Respects existing low resource requests (won't increase requests)
- Database pod detection via KubeBlocks labels


webhooks/admission/api/v1/pod_webhook.go
```diff
@@ -0,0 +1,406 @@
...
+
+// SetupWithManager sets up the webhook with the Manager
+func (r *PodMutator) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(&corev1.Pod{}).
+		WithDefaulter(r).
+		Complete()
+}
+
+//+kubebuilder:webhook:path=/mutate--v1-pod,mutating=true,failurePolicy=fail,sideEffects=None,groups=core,resources=pods,verbs=create;update,versions=v1,name=mpod.sealos.io,admissionReviewVersions=v1
```
cuisongliu: `mutate--v1-pod` The name is too simple. If there are more webhooks in the future, it may not be clear what it is.

use `mutate-qota-v1-pod` is better?

webhooks/admission/api/v1/pod_webhook.go
```diff
@@ -0,0 +1,406 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	// Labels to identify KubeBlocks database pods
+	KubeBlocksManagedByLabel = "app.kubernetes.io/managed-by"
```
cuisongliu: It's better not to use this label. It's more common. I'm not sure whether to use helm to install this value is kubeblocks or helm.

zijiren233: When using helm, this will be Helm, and the purpose of this label is to obtain ratios; for example, the ratio for ordinary Pods is 1:10, while for kubeblock it is 1:5.

PR #6040, opened 2025-09-25
feat(launchpad): add domain challenge authentication endpoints

zjy365:


frontend/providers/applaunchpad/src/pages/api/.well-known/applaunchpad-domain-challenge/[token].ts
```diff
@@ -0,0 +1,98 @@
+import type { NextApiRequest, NextApiResponse } from 'next';
+import crypto from 'crypto';
+import { jsonRes } from '../../../../services/backend/response';
```
Comment (mlhiter) [inline frontend/providers/applaunchpad/src/pages/api/.well-known/applaunchpad-domain-challenge/[token].ts:3]: there

frontend/providers/devbox/types/params.ts
```diff
@@ -10,3 +10,12 @@ export interface TemplateCallbackParams {
   templateId: string;
   [key: string]: string; // 其他可能的参数
```
mlhiter: there

PR #6049, opened 2025-09-27
fix(lifecycle): Avoid map data races for env cache when getHostEnvInCache

szediktam:
sealos panics because of ****concurrent map read and map write*** when running multi applications.
Avoid map data races for env cache when getHostEnvInCache.


cuisongliu:
Good job. Thanks your contributors !  @szediktam

PR #6181, opened 2025-11-13
feat(vlogs): add vlogs keywords escape.

bearslyricattack:
To prevent malicious injection of log queries, the enhancement of query validation through escape.

### DIFF
--- a/service/vlogs/server/query.go
+++ b/service/vlogs/server/query.go
@@ -32,7 +32,6 @@ func (v *VLogsQuery) getQuery(req *api.VlogsRequest) (string, error) {
 	v.generateDropQuery()
 	v.generateNumberQuery(req)
 	v.generateDecolorQuery()
-	fmt.Println(v.query)
 	return v.query, nil
 }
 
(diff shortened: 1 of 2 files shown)


service/vlogs/server/query.go
```diff
@@ -30,27 +31,39 @@ func (v *VLogsQuery) getQuery(req *api.VlogsRequest) (string, error) {
 	}
 	v.generateDropQuery()
 	v.generateNumberQuery(req)
+	v.generateDecolorQuery()
+	fmt.Println(v.query)
```
zijiren233: Do not continuously output sensitive information.

service/vlogs/server/query.go
```diff
@@ -160,6 +189,11 @@ var allowedNumberLevels = map[string]struct{}{
 	"s": {},
 }
 
+func isAllDigits(s string) bool {
+	_, err := strconv.Atoi(s)
+	return err != nil
+}
```
zijiren233: Function name does not match implementation.

bearslyricattack:
> Can you add a few simple `EscapeSingleQuoted` tests?

Already added.

PR #6186, opened 2025-11-14
feat(applaunchpad): bypass dns caches while challenging domains

Nixieboluo:
Signed-off-by: Nixieboluo <me@sagirii.me>

### DIFF
--- a/frontend/providers/applaunchpad/__tests__/unit/services/dns-resolver.fixture.ts
+++ b/frontend/providers/applaunchpad/__tests__/unit/services/dns-resolver.fixture.ts
@@ -38,904 +38,71 @@ function queryKey(domain: string, type: string): string {
   return `${domain}:${type}`;
 }
 
-// Mock DNS responses
-const responses = {
-  // example.org responses
-  exampleOrgA: {
-    id: 24537,
+// Helper function to create DNS response packet
+function createDnsResponse(
+  id: number,
+  questions: Array<{ name: string; type: string; class?: string }>,
+  answers: Array<{
+    name: string;
+    type: string;
+    ttl?: number;
+    class?: string;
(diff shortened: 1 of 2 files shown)


frontend/providers/applaunchpad/src/pages/api/platform/authDomainChallenge.ts
```diff
@@ -25,22 +27,62 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
+      return null;
+    });
+
     const token = nanoid();
 
-    const challengeUrl = `http://${customDomain}/api/.well-known/applaunchpad-domain-challenge/${token}`;
+    const challengeUrls = [ip4?.data, ip6?.data, customDomain].flatMap((host) => {
+      if (!host) return [];
+      return [
+        `https://${host}/api/.well-known/applaunchpad-domain-challenge/${token}`,
```
zijiren233: There's no need to use HTTPS, as most HTTPS certificates do not include IP addresses.

frontend/providers/applaunchpad/src/services/dns-resolver.ts
```diff
@@ -0,0 +1,714 @@
...
+
+  // All NS failed
+  return null;
+};
+
+/**
+ * Get authoritative NS servers of the given domain name.
+ * @returns a list of NS servers
+ */
+export const getAuthoritativeNs = async (domain: string): Promise<NSInfo | null> => {
```
zijiren233: I think there's no need to recursively resolve the `NS` records for domains here, since `NS` records rarely change, so we can directly attempt recursive resolution from the default DNS server.

zjy365 (approved):

PR #6257, opened 2025-11-27
feat(sshgate): add devbox ssh gateway service

zijiren233:


dinoallo (changes requested):
Why use `ssh.InsecureIgnoreHostKey()`?

service/sshgate/gateway/agent.go
```diff
@@ -0,0 +1,220 @@
...
+
+				if req.WantReply {
+					_ = req.Reply(true, nil)
+				}
+
+				// CRITICAL: In bastion host mode, we need to actively create
+				// an agent channel to the client immediately!
+				result.AgentChannel = g.createAgentChannelToClient(ctx)
+
+				result.CachedRequests = append(result.CachedRequests, req)
```
dinoallo, line 150: Should `MaxCachedRequests` be checked here?

service/sshgate/gateway/agent.go
```diff
@@ -0,0 +1,220 @@
...
+) (*ssh.Client, error) {
+	backendAddr := fmt.Sprintf("%s:%d", ctx.info.PodIP, g.options.SSHBackendPort)
+
+	agentClient := agent.NewClient(agentChannel)
+
+	backendConfig := &ssh.ClientConfig{
+		User: ctx.realUser,
+		Auth: []ssh.AuthMethod{ssh.PublicKeysCallback(agentClient.Signers)},
+		//nolint:gosec
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
```
dinoallo, line 205: Why use `InsecureIgnoreHostKey`?

service/sshgate/gateway/agent.go
zijiren233: No check is needed here because this request must be cached.

service/sshgate/gateway/agent.go
```diff
@@ -0,0 +1,220 @@
...
+
+		return
+	}
+
+	// Connect to backend with agent authentication if available
+	backendConn, err := g.connectToBackend(ctx, sessionResult.AgentChannel)
+	// close agent channel
+	_ = sessionResult.AgentChannel.Close()
+
+	if err != nil {
```
dinoallo, line 46: This error should be handled immediately after `connectToBackend` fails?

service/sshgate/gateway/agent.go
zijiren233: Since the purpose of the host key is solely to alert users when the server’s public key changes—thereby helping detect potential man-in-the-middle attacks—and since our scenario involves connecting over a trusted internal network to a known IP address, it is safe to ignore the devbox host key.
Moreover, users will only receive the host key generated by sshgate to verify whether a man-in-the-middle attack has occurred; for details, refer to the logic in hostkey.go.

service/sshgate/gateway/agent.go
zijiren233: This is intentional: AgentChannel is used for login, and it must be closed immediately regardless of whether the login succeeds or fails, as AgentChannel serves no purpose afterward.

zijiren233:
> Where are there so many unhandled errors?

Insignificant network I/O errors can be safely ignored, as their occurrence will not lead to catastrophic consequences.

dinoallo (approved):

PR #6283, opened 2025-12-03
feat(httpgate): add devbox http gateway

zijiren233:


service-rs/httpgate/src/proxy.rs
```diff
@@ -0,0 +1,522 @@
...
+
+#[async_trait]
+impl ProxyHttp for DevboxProxy {
+    type CTX = Option<ProxyCtx>;
+
+    fn new_ctx(&self) -> Self::CTX {
+        None
+    }
+
+    async fn request_filter(&self, session: &mut Session, ctx: &mut Self::CTX) -> Result<bool> {
```
dinoallo, line 204: Nit: If somehow the implementation of `resolve_backend` changes and it takes a long time, `request_filter` might not yield to the executor.

service-rs/httpgate/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+    init_logging(&config.log_level);
+
+    info!(listen_addr = %config.listen_addr, "Starting httpgate");
+
+    // Create shared registry
+    let registry = Arc::new(DevboxRegistry::new());
+
+    // Create Pingora server
+    let opt = Opt::default();
+    let mut server = Server::new(Some(opt)).unwrap();
```
dinoallo: Maybe log something when `unwrap` results in panic?

service-rs/httpgate/src/proxy.rs
zijiren233: Using `runtime::current_handle::spawn` to handle `handle_event -> process_new_http` in pingroa should not cause blocking.
I’m not sure which documentation or comments I missed—do you have more specific information?

PR #6346, opened 2025-12-16
feat(rauth): devbox registry auth server

zijiren233:


zzjin:
应该要放到 devbox 的 registry 部署的文件夹和逻辑里面去一起好点

---
It should be better to put it in the registry deployment folder and logic of devbox

PR #6434, opened 2025-12-24
feat(account-controller): remove payment reconcile & use Background deletion to avoid blocking issues for backup schedules

bxy4543:

### DIFF
--- a/controllers/account/controllers/namespace_controller.go
+++ b/controllers/account/controllers/namespace_controller.go
@@ -11,11 +11,12 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


controllers/account/controllers/namespace_controller.go
```diff
@@ -2750,7 +2751,9 @@ func deleteResourceAndWait(
 	namespace, name string,
 ) error {
 	deletePolicy := v12.DeletePropagationForeground // Foreground deletion, wait for child resources
-
+	if gvr.Resource == "backups" {
```
zijiren233, line 2754: The `deleteBackup` function needs to be handled, as the backup processing logic was previously moved out.

zijiren233 (approved):

PR #6704, opened 2026-02-25
feat(helm): add Helm values management for Sealos module with split configuration

cuisongliu:
…onfiguration

### DIFF
--- a/proposal/design/en/helm-values.md
+++ b/proposal/design/en/helm-values.md
@@ -1,32 +1,33 @@
-# Sealos 模块 Helm Values 配置管理
+# Helm Values Management for Sealos Modules
 
-## 目的
+## Purpose
 
-本文档用于解决模块部署后的增量配置更新问题。  
-当前 Helm 改造已基本完成，但当所有配置都放在同一个 `values.yaml` 时，运行期配置修改仍不够友好。
+This document proposes a way to support incremental configuration updates after a module is deployed.
+The current Helm refactor solves most deployment concerns, but updating runtime configuration is still difficult when everything lives in one `values.yaml`.
 
-## 问题
+## Problem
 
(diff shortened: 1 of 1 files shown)


proposal/design/zh/helm-values.md
```diff
@@ -0,0 +1,479 @@
...
+  className: "nginx"
+  annotations:
+    nginx.ingress.kubernetes.io/ssl-redirect: "true"
+    nginx.ingress.kubernetes.io/backend-protocol: "HTTP"
+  # Will be auto-configured from desktopConfig.cloudDomain
+  hosts: []
+  tls: []
+
+```
+desktop-values.yaml 主要是一些可变参数
```
fanux: There are still a lot of parameters here that could be moved to values.yaml. I suggest only keeping the 'must-change' items in desktop-values.yaml. For everything else, we should follow 'convention over configuration' and keep them in values.yaml (e.g., resources, probes). These can usually be standardized and rarely need to be touched.

proposal/design/zh/helm-values.md
```diff
@@ -0,0 +1,479 @@
...
+
+tolerations: []
+
+affinity: {}
+
+```
+
+## 脚本调整
+
+在 desktop-frontend-entrypoint.sh 把可变的values进行持久化到/root/.sealos/cloud/values/core 目录
```
bxy4543: Change the directory from values/core to core/values

proposal/design/zh/helm-values.md
cuisongliu: Would it be more suitable to keep one copy of each, so that adjustments can be easily made during changes, such as adjusting CPU, memory, and the number of replicas (which we are more likely to need to adjust in the production environment)?

PR #6791, opened 2026-03-11
feat(desktop): validate kubeconfig on desktop load 

zjy365:
After kubeconfig rotation, other sessions holding the old kubeconfig are not kicked out. Add a validation check that uses the session's kubeconfig to call k8s API on desktop load. If the old token is revoked (401), the global axios interceptor automatically clears the session and redirects to the signin page.


mlhiter:
tested?  @zjy365

PR #6822, opened 2026-03-19
fix(dbprovider): use cluster-uid label for backup queries to prevent …

aimeritething:
…name collision

Switch backup listing from app.kubernetes.io/instance (db name) to dataprotection.kubeblocks.io/cluster-uid (db UID) so that recreated databases with the same name no longer inherit old backups. Adds UUID validation, uses BackupClusterUidLabel constant, wraps cluster lookups in try/catch for proper 404 handling, and fixes query key in BackupTable.

### DIFF
--- a/frontend/providers/dbprovider/src/pages/api/backup/getBackupList.ts
+++ b/frontend/providers/dbprovider/src/pages/api/backup/getBackupList.ts
@@ -3,16 +3,10 @@ import { ApiResp } from '@/services/kubernet';
 import { authSession } from '@/services/backend/auth';
 import { getK8s } from '@/services/backend/kubernetes';
 import { jsonRes } from '@/services/backend/response';
-import { BackupClusterUidLabel } from '@/constants/db';
-
-export type Props = {
-  dbUid: string;
-};
-
-const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+import { BackupClusterUidLabel, UUID_REGEX } from '@/constants/db';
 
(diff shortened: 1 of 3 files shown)


frontend/providers/dbprovider/src/pages/api/backup/getBackupList.ts
```diff
@@ -3,25 +3,27 @@ import { ApiResp } from '@/services/kubernet';
...
 import { jsonRes } from '@/services/backend/response';
-import { crLabelKey } from '@/constants/db';
+import { BackupClusterUidLabel } from '@/constants/db';
 
 export type Props = {
-  dbName: string;
+  dbUid: string;
 };
 
+const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
```
Nixieboluo: Consider move to `constants.ts`?

frontend/providers/dbprovider/src/pages/api/v2alpha/databases/[databaseName]/backups/index.ts
```diff
@@ -64,6 +64,42 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
+      } catch (err: any) {
+        if (err?.response?.statusCode === 404) {
+          return sendError(res, {
+            status: 404,
+            type: ErrorType.RESOURCE_ERROR,
+            code: ErrorCode.NOT_FOUND,
+            message: 'Database not found.'
+          });
+        }
+        throw err;
```
Comment (Nixieboluo) [inline frontend/providers/dbprovider/src/pages/api/v2alpha/databases/[databaseName]/backups/index.ts]: Can we re-throw the original error here?

Maybe wrap with our `sendError()` (and include the original error in `cause`) is a better choice.

PR #6831, opened 2026-03-24
fix(applaunchpad): align within_5_minutes locale key

Che-Zhu:
Fix the locale key mismatch for the pod log time-range option in applaunchpad.

This aligns the zh/en locale keys with `within_5_minutes` used by the log modal.

Related issue: N/A


norberia:
Closing this PR because upstream/main has since been updated to use `within_5_minute` consistently in both LogsModal.tsx and the applaunchpad locale files as part of #6656. That makes this change outdated, and rebasing it now would only reintroduce a conflicting key shape.

PR #6896, opened 2026-04-20
chore(frontend): bump to node 26

Nixieboluo:
Use Node.js ~26.3.0~ **26.5.0** as our runtime.

Also remove `jsonwebtoken` from our deps, in favor of the new `jose` based JWT helpers.


Nixieboluo:
Let's wait for [nodejs#62526](https://github.com/nodejs/node/pull/62526).

PR #6957, opened 2026-05-13
fix(costcenter): await post-payment refetch and surface errors

calebcgates:
## Summary

In `frontend/providers/costcenter/src/pages/plan/index.tsx`, both Stripe-success handlers fire five `queryClient.invalidateQueries(...)` calls without awaiting them and without `throwOnError`, then immediately open the `CongratulationsModal`. If any of the post-payment refetches fail (transient 5xx from the cost-center backend, network blip), react-query v4 silently swallows the rejection in its internal `.catch(noop)` — the user sees the "Congratulations on your upgrade" modal while the underlying subscription/balance data still shows pre-payment state. They paid Stripe but the UI claims they're still on the previous plan — exactly the kind of bug that generates support tickets.

This PR extracts the five `invalidateQueries` calls into a `refetchAfterPayment()` helper that:

1. Calls all five invalidations in parallel via `Promise.all`.
2. Passes `{ throwOnError: true }` (react-query v4 option) so failed refetches reject the promise rather than being swallowed.
3. Catches and surfaces failures via the existing `useCustomToast` hook so the user knows to refresh.

The same five-line block appears at two call sites in this file:

- The `useEffect` Stripe-callback handler (L197-201 in main) — handles the case where the desktop wrapper redirects back with `?stripeState=success&payId=...`. Wrapped in a `void (async () => {...})()` IIFE since the useEffect callback itself must remain synchronous.
- The `PlanConfirmationModal`'s `onPaymentSuccess={async () => {...}}` handler (L885-889 in main) — handles the case where the modal stays open during checkout. Already async, so `await refetchAfterPayment()` works directly.

Both now use the new helper.

## Why

The TanStack docs describe `invalidateQueries` as fire-and-forget by default. From `@tanstack/query-core@4.35.3/build/lib/queryClient.js` (which is what costcenter pins):

```js
let promise = Promise.all(promises).then(utils.noop);
if (!(options != null && options.throwOnError)) {
  promise = promise.catch(utils.noop);  // <-- swallows rejection silently
}
return promise;
```

Without `throwOnError: true`, the refetch promise is `.catch(noop)`-wrapped before being returned. There's no observable failure signal at the call site, and the costcenter app's `QueryClient` (in `pages/_app.tsx`) doesn't configure a global `onError` handler that would catch it elsewhere — so a backend 5xx during the post-payment refetch becomes invisible to the user, who has already been redirected from Stripe back to the in-app success flow.

The choice of `await` + `throwOnError: true` + `try/catch` + `useCustomToast` is consistent with the surrounding costcenter app:

- `useCustomToast` is already imported and used in `pages/plan/index.tsx` (e.g. for transfer/recharge failures elsewhere in the file).
- The toast title is wrapped in `t('common:plan.refetch_after_payment_failed', '<English fallback>')` — the fallback string pattern is established in this app and means the PR doesn't depend on adding new i18n keys to ship. If reviewers want the i18n keys added now, happy to follow up; otherwise, the fallback works for all locales until the keys are added in `public/locales/{en,zh,zh-Hans}/common.json`.

## Scope

Six other `useQuery` violations exist in the same file (background-refetch failures on `getSubscriptionInfo`, `getLastTransaction`, etc.) — these are real but less impactful (they fail invisibly during normal page navigation rather than directly after a paid conversion). Deliberately out of scope for this PR per the "one violation type per PR" convention. Happy to follow up.

## Testing

New unit test at `frontend/providers/costcenter/__tests__/unit/refetch-after-payment.test.ts`. This is the first vitest unit test in the costcenter app — `__tests__/unit/.gitkeep` was previously the only file there. The vitest 'unit' project is already configured in `vitest.config.mts`.

Three cases:

1. All five `invalidateQueries` resolve → no toast fires, all five called with `{ throwOnError: true }`.
2. One of the five `invalidateQueries` rejects → toast fires exactly once with the failure message.
3. Regression test for pre-fix silent-failure behavior (full rejection mock → toast must fire).

```
$ cd frontend/providers/costcenter && pnpm test:ci
  ✓ unit __tests__/unit/refetch-after-payment.test.ts (3 tests) 5ms
  Test Files  1 passed (1)
       Tests  3 passed (3)
```

No existing CI integration for vitest on this app yet (the `frontends.yml` workflow only builds Docker images), so this PR doesn't affect gating. The test runs locally for any contributor who picks up the file.

Also verified:

- `pnpm lint` (`next lint`) — no new warnings on `plan/index.tsx`.
- `pnpm build` (`next build`) — production build succeeds, `/plan` route compiled.
- `tsc --noEmit` against the changed file produces zero errors (only pre-existing repo-wide SVG-asset module-resolution errors remain, which Next.js's webpack resolves at build time).
- Local pnpm matches the repo's `packageManager: pnpm@8.9.0` pin via corepack.

## DCO

Commit signed off per CONTRIBUTING.md § "Commit your changes ... commit with sign-off" (L114-119):

```
Signed-off-by: Caleb CGates <calebcgates@gmail.com>
```

### DIFF
--- a/frontend/providers/costcenter/__tests__/unit/refetch-after-payment.test.ts
+++ b/frontend/providers/costcenter/__tests__/unit/refetch-after-payment.test.ts
@@ -1,0 +1,104 @@
+// Verifies that post-payment query invalidation in pages/plan/index.tsx
+// is awaited and surfaces failures via the toast hook, instead of silently
+// leaving the UI in a stale state after Stripe payment success.
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { QueryClient } from '@tanstack/react-query';
+
+// The five queryKeys the post-payment refetch invalidates, in the same order
+// as pages/plan/index.tsx so a regression in either drifts the test.
+const POST_PAYMENT_QUERY_KEYS = [
+  ['subscription-info'],
+  ['last-transaction'],
+  ['upgrade-amount'],
+  ['card-info'],
+  ['payment-waiting-transaction']
+] as const;
+
+describe('refetchAfterPayment', () => {
+  let queryClient: QueryClient;
+  let toast: ReturnType<typeof vi.fn>;
+
+  beforeEach(() => {
+    queryClient = new QueryClient({
+      defaultOptions: { queries: { retry: false } }
+    });
+    toast = vi.fn();
+  });
+
+  // Inline replica of the helper defined in pages/plan/index.tsx so this test
+  // can exercise the behaviour without rendering the full Plan component
+  // (which pulls in next-i18next, chakra, stripe, etc.). Keep the body in
+  // sync with the implementation in pages/plan/index.tsx.
+  function makeRefetchAfterPayment(qc: QueryClient, toastFn: typeof toast) {
+    return async () => {
+      try {
+        await Promise.all([
+          qc.invalidateQueries({ queryKey: ['subscription-info'] }, { throwOnError: true }),
+          qc.invalidateQueries({ queryKey: ['last-transaction'] }, { throwOnError: true }),
+          qc.invalidateQueries({ queryKey: ['upgrade-amount'] }, { throwOnError: true }),
+          qc.invalidateQueries({ queryKey: ['card-info'] }, { throwOnError: true }),
+          qc.invalidateQueries(
+            { queryKey: ['payment-waiting-transaction'] },
+            { throwOnError: true }
+          )
+        ]);
+      } catch (err) {
+        toastFn({
+          status: 'error',
+          title:
+            'Payment succeeded, but refreshing subscription data failed. Please refresh the page.'
+        });
+      }
+    };
+  }
+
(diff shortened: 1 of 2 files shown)


calebcgates:
The CLA Page will not load for me. I've tried multiple times.

zjy365 (approved):

zjy365:
> The CLA Page will not load for me. I've tried multiple times.

Thanks! This PR is approved on my side now, only the CLA check is still pending. Could you try signing it again, or trigger the CLA recheck? Once it passes, I’ll merge it.

calebcgates:
The CLA loaded in 5 seconds, asked me to login to Github, and said it was signed and redirected me back here. I believe it's waiting for your team to assign a milestone.
Thank you @zjy365

zjy365:
> The CLA loaded in 5 seconds, asked me to login to Github, and said it was signed and redirected me back here. I believe it's waiting for your team to assign a milestone. Thank you @zjy365

Thanks! I’ve merged it.

PR #6989, opened 2026-06-08
fix(applaunchpad): harden domain challenge SSRF checks

zjy365:
## Summary

- require an authenticated kubeconfig before running AppLaunchpad domain challenge checks
- restrict custom domains to public hostnames and block private/reserved DNS targets before fetch
- prevent redirect-based SSRF and remove challenge response body/signature material leaks
- add unit coverage for auth, fragment/path input, private IPs, redirects, IPv6 handling, and response redaction

## Root cause

`/api/platform/authDomainChallenge` accepted unauthenticated `customDomain` input, constructed an outbound URL from it, followed redirects by default, and reflected invalid JSON challenge bodies. That allowed attackers to drive backend requests toward internal services and read JSON responses.

## Verification

- `pnpm --filter applaunchpad exec tsc --noEmit`
- `pnpm --filter applaunchpad test:unit`
- `pnpm --filter applaunchpad exec prettier --check src/pages/api/platform/authDomainChallenge.ts __tests__/unit/pages/api/platform/authDomainChallenge.test.ts`
- `git diff --check`

### DIFF
--- a/frontend/providers/applaunchpad/src/pages/api/platform/authDomainChallenge.ts
+++ b/frontend/providers/applaunchpad/src/pages/api/platform/authDomainChallenge.ts
@@ -90,7 +90,7 @@ const isPrivateIpv4 = (ip: string) => {
   const bytes = parseIpv4(ip);
   if (!bytes) return true;
 
-  const [first, second] = bytes;
+  const [first, second, third] = bytes;
   return (
     first === 0 ||
     first === 10 ||
@@ -98,9 +98,12 @@ const isPrivateIpv4 = (ip: string) => {
     (first === 100 && second >= 64 && second <= 127) ||
     (first === 169 && second === 254) ||
     (first === 172 && second >= 16 && second <= 31) ||
-    (first === 192 && second === 0) ||
+    (first === 192 && second === 0 && third === 0) ||
+    (first === 192 && second === 0 && third === 2) ||
     (first === 192 && second === 168) ||
     (first === 198 && (second === 18 || second === 19)) ||
+    (first === 198 && second === 51 && third === 100) ||
+    (first === 203 && second === 0 && third === 113) ||
     first >= 224
   );
 };
@@ -193,6 +196,23 @@ const authenticateRequest = async (req: NextApiRequest) => {
   );
 };
 
+const getAuthErrorCode = (error: any) => {
+  if (error === 'unAuthorization') return ResponseCode.UNAUTHORIZED;
+
+  const statusCode =
+    error?.body?.code ||
+    error?.response?.status ||
+    error?.response?.statusCode ||
+    error?.status ||
+    error?.statusCode;
+
+  if (statusCode === ResponseCode.UNAUTHORIZED || statusCode === ResponseCode.FORBIDDEN) {
+    return statusCode;
+  }
+
+  return null;
+};
+
 function isTimestampValid(timestamp: number, maxAge: number = 600): boolean {
   const now = Math.floor(Date.now() / 1000);
   const age = now - timestamp;
@@ -201,15 +221,22 @@ function isTimestampValid(timestamp: number, maxAge: number = 600): boolean {
 
 export default async function handler(req: NextApiRequest, res: NextApiResponse) {
   try {
-    const authenticated = await authenticateRequest(req)
-      .then(() => true)
-      .catch(() => false);
-    if (!authenticated) {
+    try {
+      await authenticateRequest(req);
+    } catch (authError) {
+      const authErrorCode = getAuthErrorCode(authError);
+      if (!authErrorCode) {
+        throw authError;
+      }
+
       return jsonRes(res, {
-        code: ResponseCode.UNAUTHORIZED,
+        code: authErrorCode,
         error: {
-          code: 'UNAUTHORIZED',
-          message: 'Authentication required'
+          code: authErrorCode === ResponseCode.FORBIDDEN ? 'FORBIDDEN' : 'UNAUTHORIZED',
+          message:
+            authErrorCode === ResponseCode.FORBIDDEN
+              ? 'Insufficient permissions'
+              : 'Authentication required'
         }
       });
     }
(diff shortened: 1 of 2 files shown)


frontend/providers/applaunchpad/src/pages/api/platform/authDomainChallenge.ts
```diff
@@ -4,13 +4,195 @@ import { customAlphabet } from 'nanoid';
...
+    first === 10 ||
+    first === 127 ||
+    (first === 100 && second >= 64 && second <= 127) ||
+    (first === 169 && second === 254) ||
+    (first === 172 && second >= 16 && second <= 31) ||
+    (first === 192 && second === 0) ||
+    (first === 192 && second === 168) ||
+    (first === 198 && (second === 18 || second === 19)) ||
+    first >= 224
+  );
```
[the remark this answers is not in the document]
zjy365: Fixed in 626ace7e3: narrowed 192.0.0.0 handling to specific special-use ranges and added coverage proving 192.0.1.1 remains public.

frontend/providers/applaunchpad/src/pages/api/platform/authDomainChallenge.ts
```diff
@@ -19,6 +201,19 @@ function isTimestampValid(timestamp: number, maxAge: number = 600): boolean {
...
+      .catch(() => false);
+    if (!authenticated) {
+      return jsonRes(res, {
+        code: ResponseCode.UNAUTHORIZED,
+        error: {
+          code: 'UNAUTHORIZED',
+          message: 'Authentication required'
+        }
+      });
+    }
```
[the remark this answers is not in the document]
zjy365: Fixed in 626ace7e3: known auth/permission failures now preserve 401/403, while unexpected K8s/API errors bubble to the 500 path with regression coverage.

frontend/providers/applaunchpad/src/pages/api/platform/authDomainChallenge.ts
```diff
@@ -55,8 +311,9 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
       // Make HTTP request to user's domain
       response = await fetch(url, {
         method: 'GET',
+        redirect: 'manual',
```
zijiren233: No handling of manual redirect validation and following was observed. https://undici.nodejs.org/#/?id=manual-redirect

zijiren233 (changes requested):

frontend/providers/applaunchpad/src/pages/api/platform/authDomainChallenge.ts
zjy365: Handled in 86dfc16ae. I kept redirects manual, but now explicitly allow the narrow common case: HTTP -> HTTPS upgrade only when the Location stays on the same verified domain and same challenge path. Before following, the domain is resolved again and must still resolve only to public addresses; the HTTPS request is then made to the re-validated public IP with Host/SNI set to the verified domain. Cross-host, path-changing, non-HTTPS, credentialed, non-443, and private re-resolution redirects still fail.

zjy365: Updated in 4d45be3c5: the HTTP and HTTPS challenge requests now reuse the same initially resolved and public-validated IP. The HTTPS redirect path no longer performs another DNS lookup; it connects to the original resolved IP with Host and TLS SNI set to the verified domain. Unsafe redirects still fail.

frontend/providers/applaunchpad/src/pages/api/platform/authDomainChallenge.ts
```diff
@@ -4,13 +4,361 @@ import { customAlphabet } from 'nanoid';
...
+
+  return redirectUrl.toString();
+};
+
+const fetchChallengeUrl = (url: string, verifiedDomain: string) => {
+  return fetch(url, {
+    method: 'GET',
+    redirect: 'manual',
+    headers: {
+      Host: verifiedDomain,
```
zijiren233: Node.js’s `fetch` cannot set the `Host` header:
```bash
docker run --rm -i node:20.20.2-alpine node - <<'NODE'
const http = require('http');

const server = http.createServer((req, res) => {
  res.end(JSON.stringify({ host: req.headers.host }));
});

server.listen(0, '127.0.0.1', async () => {
  const { port } = server.address();

  try {
    const res = await fetch(`http://127.0.0.1:${port}/challenge`, {
      headers: {
        Host: 'example.com',
        'User-Agent': 'Sealos-AppLaunchpad-Domain-Verifier/1.0'
      }
    });

    console.log(await res.text());
  } finally {
    server.close();
  }
});
NODE
```

zijiren233: You should implement it similarly to the HTTPS implementation, using `http.request`。

zijiren233 (changes requested):

frontend/providers/applaunchpad/src/pages/api/platform/authDomainChallenge.ts
zjy365: Fixed in d71376445. The HTTP challenge path now uses http.request instead of fetch, so it connects to the resolved IP while explicitly sending Host: verifiedDomain. The tests were updated to mock http.request directly and assert the HTTP request hostname is the resolved IP and the Host header is the verified domain; the HTTPS redirect path still reuses the same resolved IP with SNI/Host set to the verified domain.

zjy365: Done in d71376445. I replaced the HTTP fetch path with http.request, matching the HTTPS implementation style. HTTP now connects to the resolved IP and sends Host: verifiedDomain explicitly; HTTPS redirect still reuses the same resolved IP with Host/SNI set to the verified domain. Added tests now assert http.request receives the resolved IP as hostname and the verified domain as Host.

zijiren233 (approved):

Nixieboluo (approved):
lgtm

PR #6998, opened 2026-06-13
feat(applaunchpad): improve public domain and image port flows

mlhiter:
## Summary

This PR improves App Launchpad create/edit networking behavior for the release-v5.1 branch.

- Detect image exposed ports and use them as guarded defaults without overwriting later manual edits.
- Make managed public-domain prefixes editable with local format/reserved-prefix validation, same-form duplicate detection, availability precheck, and submit-time fallback validation.
- Translate public-domain admission conflicts into clear product errors, including safe same-workspace owner attribution and generic cross-workspace conflicts.
- Add App Launchpad project docs/runbook updates and focused helper coverage for public-domain and image-port logic.

## Verification

- `pnpm --filter ./providers/applaunchpad exec tsc --noEmit --pretty false`
- `pnpm --filter ./providers/applaunchpad lint` (passes with existing react-hooks warnings)
- `git diff --check release-v5.1...HEAD`
- `git diff --check`
- `pnpm dlx tsx@latest --tsconfig=tsconfig.json -e "...public-domain helper checks..."`
- `pnpm dlx tsx@latest --tsconfig=tsconfig.json -e "...image exposed port helper checks..."`

Note: direct `vitest run` is currently blocked because App Launchpad does not have Vitest alias config for `@/*`; the helper behavior above was spot-checked through `tsx --tsconfig`.

## Base

Target branch: `release-v5.1`.


frontend/providers/applaunchpad/AGENTS.md
```diff
@@ -0,0 +1,80 @@
+# App Launchpad Agent Notes
```
zjy365: This file can be deleted.

zjy365 (changes requested):

frontend/providers/applaunchpad/src/utils/image-exposed-ports.ts
```diff
@@ -0,0 +1,440 @@
...
+    Accept: ACCEPT_HEADER
+  };
+
+  if (auth?.username && auth.password && image.registry !== DEFAULT_REGISTRY) {
+    headers.Authorization = `Basic ${Buffer.from(`${auth.username}:${auth.password}`).toString(
+      'base64'
+    )}`;
+  }
+
+  let response = await fetch(url, {
```
zjy365: P1: This SSRF guard is still bypassable. `assertPublicRegistryHost()` validates the registry host before the request, but `fetch()` follows redirects by default and will also perform its own DNS resolution at connection time. An attacker-controlled public registry can pass the initial lookup, then return a 302 to an internal/metadata address, or use DNS rebinding so the actual fetch connects to a private IP. Please set registry/auth/config fetches to `redirect: "manual"` and reject or re-validate every `Location`; also enforce the same public-host/IP checks at connection time, for example via an undici dispatcher/custom lookup or an equivalent allowlisted proxy.

zjy365 (approved):

PR #6999, opened 2026-06-13
feat(template): support managed template repository layout

mlhiter:
## What this PR does

This PR adds Template App support for the managed template repository layout produced by Sealos Admin.

- Reads category metadata from `config/categories.json` in the cloned template repository, with the existing env fallback kept server-side.
- Serves repository-local README/icon assets through `/api/templateAsset` so private or offline stores do not depend on public raw URLs.
- Supports split template manifests such as `template/<name>/index.yaml` plus `template/<name>/manifests/*.yaml` while keeping legacy single-file templates compatible.

## Why this is needed

The paired Admin template-management flow writes templates, categories, README files, icons, and manifests into an internal Git/Gogs repository. Template App needs to consume that same repository format for the store and deploy flows.

## Test

- `pnpm -C frontend/providers/template build`

Note: the build completed successfully with existing React Hook lint warnings.

### DIFF
--- a/frontend/providers/template/src/pages/api/templateAsset.ts
+++ b/frontend/providers/template/src/pages/api/templateAsset.ts
@@ -2,6 +2,7 @@ import { jsonRes } from '@/services/backend/response';
 import { readTemplateAssetFile } from '@/utils/templateAssets';
 import type { NextApiRequest, NextApiResponse } from 'next';
 import path from 'path';
+import { ensureRepoFresh } from '@/services/backend/template-repo';
 
 const MIME_BY_EXT: Record<string, string> = {
   '.svg': 'image/svg+xml',
@@ -14,7 +15,7 @@ const MIME_BY_EXT: Record<string, string> = {
   '.txt': 'text/plain; charset=utf-8'
 };
 
-export default function handler(req: NextApiRequest, res: NextApiResponse) {
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
   const { templateName, asset } = req.query as {
     templateName?: string;
     asset?: string;
@@ -25,6 +26,8 @@ export default function handler(req: NextApiRequest, res: NextApiResponse) {
   }
 
   try {
+    await ensureRepoFresh();
+
     const originalPath = process.cwd();
     const jsonPath = path.resolve(originalPath, 'templates.json');
     const result = readTemplateAssetFile({ jsonPath, templateName, assetUrl: asset });
--- a/frontend/providers/template/src/pages/api/templateAsset.ts
+++ b/frontend/providers/template/src/pages/api/templateAsset.ts
@@ -1,0 +1,41 @@
+import { jsonRes } from '@/services/backend/response';
+import { readTemplateAssetFile } from '@/utils/templateAssets';
+import type { NextApiRequest, NextApiResponse } from 'next';
+import path from 'path';
+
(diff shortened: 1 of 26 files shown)


frontend/providers/template/src/pages/api/templateAsset.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (!templateName || !asset) {
+    return jsonRes(res, { code: 400, message: 'templateName and asset are required.' });
+  }
+
+  try {
+    await ensureRepoFresh();
+
+    const originalPath = process.cwd();
+    const jsonPath = path.resolve(originalPath, 'templates.json');
+    const result = readTemplateAssetFile({ jsonPath, templateName, assetUrl: asset });
```
zjy365, line 33: This makes the new public asset endpoint a generic file reader for the cloned template repository. `asset` comes directly from the query string and is only constrained later to stay under `templates/`, so an unauthenticated caller can request files that were never declared as this template’s README/icon, for example enough `../` segments to reach `templates/.git/config` or `/config/categories.json`. In private/offline repo deployments this can leak repo config, unpublished templates, and other non-asset files. Please authorize the asset against the template’s declared `spec.readme`/`spec.icon`/`i18n` assets, and only serve derived README images from an explicit allowlist of safe locations/extensions.

frontend/providers/template/src/utils/templateAssets.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+export function getTemplateRepoRootPath() {
+  return path.resolve(process.cwd(), 'templates');
+}
+
+function assertInRepo(repoRootPath: string, targetPath: string) {
+  const root = path.resolve(repoRootPath);
+  const target = path.resolve(targetPath);
+  if (target !== root && !target.startsWith(`${root}${path.sep}`)) {
```
zjy365, line 35: This boundary check is path-string based, so it does not protect against symlinks inside the template repository. A repo entry such as `template/foo/leak.txt -> /etc/passwd` still resolves under `templates/` here, but `fs.readFileSync()` follows the symlink and returns the target outside the repo. The split-manifest reader has the same shape when reading files from `manifests/`. Please validate real paths before reading and reject symlinks, or use `lstat` plus a realpath repo-root check and only allow regular files.

frontend/providers/template/src/services/backend/template-repo.ts
```diff
@@ -133,4 +149,33 @@ export async function updateRepo() {
 
   const jsonContent = JSON.stringify(jsonObjArr, null, 2);
   fs.writeFileSync(jsonPath, jsonContent, 'utf-8');
+  lastRepoSyncAt = Date.now();
+  clearTemplateRuntimeCaches();
+}
+
+export async function ensureRepoFresh({ force = false }: { force?: boolean } = {}) {
```
zjy365, line 156: Moving repo refresh from the existing background cron into `ensureRepoFresh()` changes the freshness model in a problematic way. With no traffic the catalog no longer updates in the background, and every read path now has to remember to call this helper before touching `templates.json`; any new or existing endpoint that misses it will silently serve stale data. It also puts `git pull` on the user request path whenever the TTL expires. Please keep a single background refresh mechanism (or another centralized scheduler) and let read handlers only read the current catalog/cache, instead of making each API endpoint responsible for repository synchronization.

PR #7005, opened 2026-06-17
fix(dbprovider): hide empty monitor metrics

zjy365:
## Summary
- hide DBProvider monitor charts that currently render empty data
- fix the malformed MySQL aborted connections PromQL query
- ignore local CodeGraph index output

## Testing
- Not run; change is limited to chart visibility and query string cleanup.


Nixieboluo (approved):
lgtm

PR #7012, opened 2026-06-22
fix(auth): preserve nickname on repeated login

zjy365:
## Summary

- stop repeated sign-ins from overwriting user-managed nicknames with provider profile names
- keep existing avatar refresh behavior for returning users
- add a unit test covering the repeated-login nickname preservation path

## Root cause

Existing users were updated on every non-password login with the latest provider profile name. That made admin-edited nicknames revert to the GitHub login, phone number, or OAuth profile name after the next login.

## Tests

- `pnpm vitest run --project unit src/__tests__/unit/backend/auth/global-auth.test.ts`
- `pnpm vitest run --project unit src/__tests__/unit/backend/auth src/__tests__/unit/backend/oauth2-idp`


Nixieboluo (approved):
lgtm

PR #7013, opened 2026-06-22
fix(applaunchpad): hydrate form quantities

Nixieboluo:
Fixes the Applaunchpad client crash caused by serialized or form-cloned `Quantity` values losing methods like `milliValue()`.

Adds runtime quantity hydration at conversion/form boundaries plus a regression test for serialized CPU, memory, and storage values.


Nixieboluo:
superseded by #7018

PR #7021, opened 2026-06-25
fix(dbprovider): apply create patch yamls

Nixieboluo:
Apply DB create-time patch YAMLs from `constants/db.ts`, starting with PostgreSQL reconfiguration to remove `bg_mon` from `shared_preload_libraries`.

Wire the shared patch flow into all dbprovider create paths and add unit coverage for the PostgreSQL OpsRequest payload.


Nixieboluo:
Not eligible.

PR #7028, opened 2026-06-25
fix(applaunchpad): preserve init container volumes during yaml patch

zjy365:
## Summary
- Preserve volumes referenced only by initContainers when patching Deployment and StatefulSet YAML.
- Add a unit regression test for Deployment patching so configMap volumes used only by initContainers are retained.

## Testing
- pnpm --filter ./providers/applaunchpad test:unit


Nixieboluo (approved):
lgtm

PR #7034, opened 2026-06-26
fix(applaunchpad): delete pvcs for owner-referenced apps

zjy365:
## Summary
- remove the `/api/delApp` ownerReferences fast path so app deletion always runs explicit dependent cleanup
- keep the existing `app=<name>` PVC collection delete for owner-referenced StatefulSet apps
- add a regression test covering a StatefulSet app with `cloud.sealos.io/owner-references=ready`

## Root Cause
`/api/delApp` assumed ownerReferences cascade deletion would clean every dependent resource and returned immediately after deleting the workload. StatefulSet `volumeClaimTemplates` PVCs are retained by default, so the PVC cleanup block was skipped and app PVCs remained after deletion.

## Validation
- `pnpm --dir frontend/providers/applaunchpad exec vitest run`

## Additional Information

Use `persistentVolumeClaimRetentionPolicy` after Kubernetes v1.32: https://github.com/labring/sealos/pull/7034#issuecomment-4807811441

### DIFF
--- a/frontend/providers/applaunchpad/__tests__/unit/pages/api/delApp.test.ts
+++ b/frontend/providers/applaunchpad/__tests__/unit/pages/api/delApp.test.ts
@@ -1,0 +1,96 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { DeleteAppByName } from '@/pages/api/delApp';
+
+const authSessionMock = vi.hoisted(() => vi.fn());
+const getK8sMock = vi.hoisted(() => vi.fn());
+
+vi.mock('@/services/backend/auth', () => ({
+  authSession: authSessionMock
+}));
+
+vi.mock('@/services/backend/kubernetes', () => ({
+  getK8s: getK8sMock
+}));
+
+function notFound() {
+  return Promise.reject({
+    body: {
+      code: 404
+    }
+  });
+}
+
+function createK8sContext() {
+  return {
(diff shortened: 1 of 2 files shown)


zjy365:
Fix approach note:

Kubernetes already has a native StatefulSet control for this behavior: `.spec.persistentVolumeClaimRetentionPolicy`, stable since Kubernetes v1.32. For Launchpad-created StatefulSets, the Kubernetes-native shape would be:

```yaml
spec:
  persistentVolumeClaimRetentionPolicy:
    whenDeleted: Delete
    whenScaled: Retain
```

That would make the StatefulSet controller delete PVCs created from `volumeClaimTemplates` when the StatefulSet is deleted.

This PR keeps the narrower 5.1 hotfix in the delete path instead: it preserves the existing ownerReferences fast path, deletes the Deployment/StatefulSet as before, and then explicitly deletes PVCs with `app=<name>` before returning. The reason is compatibility with existing apps and clusters: changing StatefulSet generation only helps newly created or updated StatefulSets that carry the retention policy, while the reported leak can happen on already-created apps whose StatefulSets do not have that field.

A more complete follow-up can add `persistentVolumeClaimRetentionPolicy.whenDeleted=Delete` to generated StatefulSets while keeping this explicit delete as a compatibility fallback for older apps / older clusters.

Nixieboluo (approved):
lgtm

PR #7043, opened 2026-07-03
feat(desktop): add core-js polyfill

Nixieboluo:
## Summary

  Adds a CoreJS polyfill to the desktop app entrypoint so the client bundle has stable JavaScript runtime APIs available in older Chromium-based browsers.

  Changes included:

  - Adds `core-js@3.33.3` as a direct `desktop` dependency
  - Imports `core-js/stable` at the top of `desktop/src/pages/_app.tsx`
  - Updates the lockfile importer entry for `desktop`

  ## Compatibility

  This targets Chrome 86 and newer.

  The polyfill is loaded before the desktop app initializes, so stable ECMAScript APIs provided by CoreJS are available across the desktop client runtime. This is intended to support
  environments based on Chromium 86, including Chrome 86 and compatible embedded Chromium/WebView runtimes.


Nixieboluo:
Closes https://github.com/labring-sigs/sealos-issues/issues/330.

zjy365 (approved):

PR #7049, opened 2026-07-08
fix(license): parse Kubernetes resource quantities

mlhiter:
## Summary

- Parse Kubernetes resource quantities in the license cluster info API instead of assuming node memory is always reported in `Ki`.
- Convert memory quantities to bytes before reporting GiB, while keeping CPU quantities in cores.
- Add regression coverage for `Ki`, `Mi`, `Gi`, and millicore CPU quantities.

## Testing

- `pnpm --filter ./providers/license run test:quantity`
- `pnpm --filter ./providers/license run build`

Notes: the build passes with existing local warnings for Node v23.11.0 vs the package requirement of 20.20.2, existing `_app.tsx` hook dependency warnings, outdated Browserslist data, and static-generation i18next initialization messages.

### DIFF
--- a/frontend/providers/license/package.json
+++ b/frontend/providers/license/package.json
@@ -7,6 +7,7 @@
     "build": "next build",
     "start": "next start",
     "lint": "next lint",
+    "test:quantity": "node src/utils/kubernetesQuantity.test.js",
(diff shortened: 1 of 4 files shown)


mlhiter:
Superseded by #7050, which targets release-v5.1.

PR #7052, opened 2026-07-10
fix(costcenter): use session region for account requests

luppyw:
## Summary
- use the verified session region as the fallback when costcenter API requests do not pass `regionUid`
- keep account balance requests aligned with the region used by the logged-in desktop session
- preserve explicit `regionUid` behavior for pages that already choose a region

## Verification
- `git diff --check`
- `git commit` pre-commit gitleaks checks passed
- `pnpm --dir frontend --filter costcenter lint` could not run locally because `frontend/node_modules` is missing and `next` is not installed in this sparse checkout

## Notes
- No related issue was provided.


luffy0x:
Temporarily closing this PR because the 70 cluster deployment has been rolled back. We'll revisit the costcenter balance fix next week.

PR #7071, opened 2026-07-20
fix(desktop): secure verification codes in MongoDB

zjy365:
## Summary

- keep the active verification-code path on MongoDB instead of introducing new Prisma tables
- limit each code to 10 atomic failed attempts and consume a successful code exactly once
- return a per-send `challengeId` and require it across login, bind, unbind, change-binding, and alert verification flows
- reset attempts and rotate both the code and challenge when resending
- add Mongo-backed identifier and IP send limits with TTL indexes and failed-send compensation
- store user/channel/scenario-bound change-binding flow tickets in MongoDB
- show remaining attempts, lock state, retry countdown, and localized rate-limit feedback
- strictly validate identifiers, six-digit codes, and UUID challenges before Mongo queries
- initialize verification indexes once per Mongo client and reuse collections within each reservation
- add focused Mongo concurrency and regression coverage

## Storage and limits

- verification code lifetime: 5 minutes
- resend interval: 60 seconds
- verification attempts: 10 per code
- identifier sending: 1/minute and 5/hour
- IP sending: 20/10 minutes

All verification counters and successful consumption use conditional single-document MongoDB operations. MongoDB TTL indexes clean up verification codes, rate-limit windows, and flow tickets.

## Validation

- `pnpm exec tsc --noEmit`
- `pnpm test:ci` (91 passed, 2 todo)
- `pnpm lint` (passed with pre-existing Hook warnings)
- `pnpm build`
- `git diff --check`
- Prettier check on all changed TypeScript, TSX, and JSON files

## Rollout requirements

This PR intentionally remains draft until the deployment path is confirmed:

- do not run old and new frontend pods together; old pods ignore `challengeId` and attempt counts, so use Recreate or drain old pods before starting the new version
- force clients to reload at deployment because verification requests now require `challengeId`
- confirm the ingress overwrites `X-Real-IP` before treating the IP limiter as an authoritative client-IP control
- normalize email verification lookup keys consistently so case-only resends invalidate the previous challenge

Change-binding and alert creation consume verification credentials before downstream SQL/provider writes. A downstream failure is fail-closed and requires the user to request another code.


dinoallo (approved):
LGTM

zjy365:
I’d like to suggest that this PR also be backported to the `release-5.1` branch. If you’re an agent working on this, please make sure this PR is included.

PR #7073, opened 2026-07-22
fix(cleanup): preserve shared project logic

mlhiter:
## Summary

This is a follow-up to #7072, which was merged before its corrective commit landed.

- Restore shared ObjectStorage metrics, query, and interface code used by retained controllers.
- Restore the generic MinIO Prometheus request path used by retained services.
- Relocate the two MinIO client constructors into the retained shared package so deleted controller code stays deleted.
- Restore the retained pnpm lockfile state and Go workspace sums.

This PR does not restore any migrated frontend provider, controller, or service directory and does not change their replacement repositories.

## Verification

- `go test -run '^$' ./account/... ./resources/... ./pkg/objectstorage/...` from `controllers`
- `go test -run '^$' ./pkg/...` from `service`
- `pnpm install --lockfile-only --frozen-lockfile --offline --ignore-scripts` from `frontend`
- Structured lockfile comparison: only the six migrated provider importers are removed; all 15 retained importers and package snapshots are unchanged
- `git diff --check upstream/release-v5.1..HEAD`


cuisongliu (approved):
CC @bxy4543

PR #7093, opened 2026-07-23
feat(desktop): open Brain by default from root

norberia:
## What changed

- centralize the root-path startup target selection
- preserve explicit app/Desktop navigation and restorable fullscreen apps ahead of the Brain default
- open `system-brain` when it is available and no higher-priority target exists
- fall back to Desktop when Brain is unavailable
- keep Desktop chrome hidden behind a dark startup mask until the initial iframe loads
- make automatic, manual, `quitGuide`, and modal guide paths honor `commonConfig.guideEnabled`
- keep real regional announcements visible when the guide fallback is disabled

## Why

The root page previously had separate app-opening and restore flows. It had no environment-aware default target, and the split initialization could expose Desktop before Brain or compete with restored app state.

This change gives the root path one deterministic resolver while retaining the existing login, guest, workspace-switch, Stripe callback, autolaunch, and refresh-restore behavior.

## Impact

- environments with Brain installed enter Brain directly from `/`
- environments without Brain continue to enter Desktop
- `?openapp=` remains an explicit Desktop entry
- explicit app links and refreshed fullscreen Template apps are not overwritten by the Brain default
- overseas environments can disable Desktop onboarding without hiding actual regional announcements

## Validation

- `pnpm --dir frontend/desktop test:ci` — 104 passed, 2 existing todo
- `pnpm --dir frontend/desktop exec tsc --noEmit`
- `pnpm --dir frontend/desktop lint` — passed with existing Hook dependency warnings
- `pnpm --dir frontend/desktop build`

Closes labring/sealos-private#48

### DIFF
--- a/frontend/desktop/src/components/desktop_content/index.tsx
+++ b/frontend/desktop/src/components/desktop_content/index.tsx
@@ -7,9 +7,9 @@ import { WindowSize } from '@/types';
 import { Box, Flex, Image, Button, Text } from '@chakra-ui/react';
 import { useTranslation } from 'next-i18next';
 import dynamic from 'next/dynamic';
-import { useRouter } from 'next/router';
 import { useCallback, useEffect, useRef, useState } from 'react';
 import { createMasterAPP, masterApp } from 'sealos-desktop-sdk/master';
+import type { InitialAppLaunchState } from '@/utils/initialAppTarget';
 import { ChakraIndicator } from './ChakraIndicator';
 // import Apps from './apps';
 import IframeWindow from './iframe_window';
@@ -39,9 +39,13 @@ export const blurBackgroundStyles = {
   borderRadius: '12px'
 };
 
-export default function Desktop() {
+type DesktopProps = {
+  initialAppLaunch: InitialAppLaunchState;
+  onInitialAppLoaded: (appKey: string) => void;
+};
+
+export default function Desktop({ initialAppLaunch, onInitialAppLoaded }: DesktopProps) {
   const { t } = useTranslation();
-  const router = useRouter();
   const { isAppBar } = useDesktopConfigStore();
   const {
     installedApps: apps,
@@ -50,8 +54,7 @@ export default function Desktop() {
     setToHighestLayerById,
     closeAppById,
     setAutoLaunch,
-    currentAppKey,
-    autolaunch
+    currentAppKey
   } = useAppStore();
   const backgroundImage = useConfigStore().layoutConfig?.backgroundImage;
   const { backgroundImage: desktopBackgroundImage } = useAppDisplayConfigStore();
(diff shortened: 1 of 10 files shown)


frontend/desktop/src/pages/index.tsx
```diff
@@ -145,29 +164,160 @@ export default function Home({ sealos_cloud_domain }: { sealos_cloud_domain: str
...
+      ) => {
+        const target = resolveInitialAppTarget({
+          installedAppKeys: state.installedApps.map((app) => app.key),
+          autolaunchAppKey: state.autolaunch,
+          hasOpenAppQuery: initialOpenAppIntent?.hasOpenAppQuery || false,
+          queryAppKey: parsedOpenApp.appkey,
+          restoreAppKeys: [
+            state.currentAppKey,
+            sessionStorage.getItem(SESSION_RESTORE_APP_KEY) || undefined
+          ],
```
zjy365, line 253: The tab-local restore key should take precedence over `state.currentAppKey`. Brain's `currentAppKey` is persisted through Zustand/localStorage and is therefore shared across tabs, while `SESSION_RESTORE_APP_KEY` is sessionStorage-backed and represents the current tab's fullscreen app. If another tab opens Brain, refreshing a tab that is currently showing a fullscreen Template App can resolve Brain first here and overwrite that tab's intended restore target. Please reverse these candidates (sessionStorage first, persisted `currentAppKey` second) and add a regression test where both are valid.

zjy365 (commented):
The main flow looks correct, but there is one restore-priority edge case to address.

frontend/desktop/src/components/desktop_content/index.tsx
```diff
@@ -90,56 +93,17 @@ export default function Desktop() {
     }
 
     sessionStorage.setItem(SESSION_RESTORE_APP_KEY, currentAppKey);
-  }, [isClient, currentAppKey, isRestoring]);
+  }, [isClient, currentAppKey, isInitialAppLoading]);
```
zjy365, line 92: The reordered resolver candidates are still defeated before the resolver reads them. On mount, `currentAppKey` can rehydrate as `system-brain` from shared localStorage; once `isClient` becomes true, this effect removes the tab-local Template restore key while the initial target is still resolving. `init()` then awaits the installed-app request, and the resolver eventually sees no Template key and selects Brain again. Please avoid mutating restore storage while `initialAppLaunch` is resolving/loading (for example, return while `isInitialAppLoading` is true), then update or clear it only after the selected target is ready. The regression test should exercise this effect timing, not only the pure resolver with a prebuilt candidate array.

frontend/desktop/src/components/desktop_content/index.tsx
```diff
@@ -213,10 +177,11 @@ export default function Desktop() {
 
   const quitGuide = useCallback(
     ({ appKey }: { appKey: string }) => {
+      if (!commonConfig?.guideEnabled) return;
```
zjy365: `guideEnabled` should gate only `openGuideModal()`, not closing the source app. `quitGuide` is emitted after the user chooses to exit a fullscreen sub-app guide; returning here leaves that app open and traps the user in it. Keep `closeDesktopApp({ appKey })` unconditional, then open the Desktop guide modal only when `guideEnabled` is true. Please add coverage for the disabled-guide event path.

zjy365 (commented):
Two behavior issues remain after rechecking the latest commit.

zjy365 (approved):

PR #7126, opened 2026-07-28
fix(account): make hourly billing reconciliation resumable and idempotent

zijiren233:
- schedule every ready billing hour from a persisted Mongo checkpoint
- retry failed owner reconciliation and preserve checkpoint progress
- use stable billing IDs with Mongo upsert semantics
- recover stable unsettled billings independently from monitor data
- make Cockroach balance and credits deductions idempotent
- reconstruct historical debt and subscription state using transaction UpdatedAt
- scope subscription history queries to active workspaces in the target hour
- add focused unit and Testcontainers runtime coverage

### DIFF
--- a/controllers/account/controllers/billing_controller.go
+++ b/controllers/account/controllers/billing_controller.go
@@ -48,8 +48,10 @@ type BillingTaskRunner struct {
 func (r *BillingTaskRunner) NeedLeaderElection() bool { return true }
 
 const (
-	billingMonitorDelay = 5 * time.Minute
-	billingRetryDelay   = time.Minute
+	billingMonitorDelay              = 5 * time.Minute
+	billingRetryDelay                = time.Minute
+	defaultBillingMaxCatchupDuration = 24 * time.Hour
+	billingMaxCatchupDurationEnv     = "BILLING_MAX_CATCHUP_DURATION"
 )
 
 var (
@@ -142,7 +144,7 @@ type BillingReconciler struct {
 	DBClient               database.Account
 	AccountV2              database.AccountV2
 	Properties             *resources.PropertyTypeLS
-	reconcileBillingFunc   func(owner string, billings []*resources.Billing) error
+	reconcileBillingFunc   func(owner string, billings []*resources.Billing, endHourTime time.Time) error
 	executeBillingHourFunc func(time.Time) error
 	concurrentLimit        int64
 	DebtUserMap            *maps.ConcurrentMap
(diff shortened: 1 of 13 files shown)


controllers/pkg/database/mongo/account.go
```diff
@@ -216,6 +217,101 @@ func (m *mongoDB) GetOwnersRecentUpdates(
...
+	filter := bson.M{
+		"owner": ownerListFilter(ownerList),
+		"time":  billingTime,
+		"type":  common.Consumption,
+		"app_type": bson.M{"$nin": []int{
+			int(resources.AppType[resources.CVM]),
+			int(resources.AppType[resources.LLMToken]),
+		}},
+	}
+	cursor, err := m.getBillingCollection().Find(context.Background(), filter)
```
dinoallo: Should we add a timeout?

controllers/pkg/database/mongo/account.go
```diff
@@ -216,6 +217,101 @@ func (m *mongoDB) GetOwnersRecentUpdates(
...
+}
+
+func (m *mongoDB) GetUnsettledBillingsAt(
+	billingTime time.Time,
+) (map[string][]*resources.Billing, error) {
+	filter := bson.M{
+		"time":     billingTime,
+		"type":     common.Consumption,
+		"status":   resources.Unsettled,
+		"order_id": primitive.Regex{Pattern: "^bh_"},
```
dinoallo, line 260: Legacy random-ID Unsettled billings are excluded here, while pendingOwnerBillings also suppresses regeneration when a legacy record has the same business key. Such records will be neither retried nor replaced. Could we include and reuse these legacy records as described in the design?

controllers/pkg/database/cockroach/accountv2.go
```diff
@@ -1066,18 +1067,33 @@ func (c *Cockroach) AddDeductionBalanceWithCredits(
...
 		userUID, dErr := c.GetUserUID(ops)
 		if dErr != nil {
 			return fmt.Errorf("failed to get user uid: %w", dErr)
 		}
 		var credits []types.Credits
-		if dErr = c.DB.Where("user_uid = ? AND expire_at > ? AND status = ?", userUID, time.Now().UTC(), types.CreditsStatusActive).Order("expire_at ASC").Find(&credits).Error; dErr != nil {
+		if dErr = tx.Where(
+			"user_uid = ? AND expire_at > ? AND status = ?",
+			userUID,
+			time.Now().UTC(),
```
dinoallo: For historical catch-up, should credit eligibility be evaluated at the billing hour instead of retry time? Using time.Now() means replaying after a credit expires may produce a different charge.

controllers/account/controllers/billing_controller.go
```diff
@@ -99,92 +139,452 @@ type BillingReconciler struct {
...
+	}
+	if !exists {
+		checkpoint = target.Add(-time.Hour)
+	}
+	for _, billingTime := range billingHoursAfter(checkpoint, target) {
+		execute := r.ExecuteBillingTaskAt
+		if r.executeBillingHourFunc != nil {
+			execute = r.executeBillingHourFunc
+		}
+		if err := execute(billingTime); err != nil {
```
dinoallo, line 204: A permanently failing owner blocks this hour and every subsequent hour indefinitely. What is the operational escape hatch for malformed records, such as owner/hour failure tracking, alerting, and controlled manual resolution?

dinoallo (commented):
Please check the comments

controllers/account/controllers/billing_controller.go
zijiren233: The checkpoint remains unchanged when any owner reconciliation fails, preserving billing completeness for the current replay window.

The operational controls added in this change are:

- `sealos_account_billing_failed_owners` reports the number of failed owners in the latest batch.
- `sealos_account_billing_reconcile_failures_total` counts failed billing-hour reconciliations.
- `sealos_account_billing_processing_started_timestamp_seconds` identifies a stuck reconciliation.
- `BILLING_MAX_CATCHUP_DURATION` bounds historical replay work. It defaults to `24h`, so a checkpoint that is several days behind is replayed only within the configured window.

The catch-up window limits recovery duration and backlog processing cost. It keeps the current failed hour subject to the normal checkpoint success requirement, so a continuously failing owner can still block progress within the active window.

controllers/pkg/database/mongo/account.go
zijiren233: This is for potential future idempotency considerations; the old quantities should be minimal, and old data is not considered.

dinoallo (approved):

bxy4543 (approved):

PR #7119, opened 2026-07-27
ci(lifecycle): lint test-only changes without e2e

dinoallo:
## Summary

- Add missing Apache license headers to lifecycle kubeconfig and cert test files.
- Keep lifecycle `golangci-lint` running for lifecycle Go test-only diffs.
- Limit PR lint reporting to issues introduced by the PR so existing lifecycle lint debt does not block unrelated header-only changes.
- Skip core and k3s lifecycle e2e workflows for lifecycle test-only diffs.

No sealos runtime behavior changes.

## Verification

- Loaded `.github/workflows/ci.yml` with PyYAML and ran `git diff --check`.
- Verified the lifecycle test-only diff resolves to the `./lifecycle` lint matrix.
- Ran lifecycle `golangci-lint` locally with `--new-from-rev=upstream/main`; it reported 0 issues.

PR #7128, opened 2026-07-28
ci(lifecycle): keep lint incremental on push

dinoallo:
## Background

After #7119 was merged, the CI workflow on `main` failed for commit `17eb3112f2d8c02abe65ccf3e35563ed662ce1dc`. The pull request path linted only new lifecycle issues, but the `push` path still ran full lifecycle lint. That exposed existing lifecycle lint debt and made the post-merge CI red.

## Changes

- Resolve lifecycle modules from `github.event.before` to `github.sha` on push events, matching the pull request scoped module behavior.
- Pass `--new-from-rev` to `golangci-lint` for push events with a valid before SHA, so post-merge lint checks only issues introduced by the merged commit range.
- Keep full lifecycle lint as the fallback for workflow events without a usable diff base.

## Verification

- Loaded `.github/workflows/ci.yml` with PyYAML.
- Ran `git diff --check -- .github/workflows/ci.yml`.
- Simulated the failed merge commit range with `scripts/resolve-modules.sh ./lifecycle --changed HEAD^ HEAD`; it resolves to `./lifecycle`.
- Ran lifecycle `golangci-lint v2.12.2 --new-from-rev=HEAD^`; it reported 0 issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,15 +55,18 @@ jobs:
       - name: Resolve Modules
         id: set-matrix
         run: |
+          BASE_REF=""
           if [[ "${{ github.event_name }}" == "pull_request" ]]; then
             BASE_REF="${{ github.event.pull_request.base.sha }}"
             HEAD_REF="${{ github.sha }}"
-          else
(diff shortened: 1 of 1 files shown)


dinoallo:
Superseded by #7130. This PR has stale failed title-check runs from its original title; #7130 uses the same fix commit with a compliant title from creation.

PR #7196, opened 2026-08-10
feat(user): issue kubeconfigs with token requests

dinoallo:
Use the ServiceAccount TokenRequest subresource instead of manually managed service-account-token Secrets, and request long-lived tokens as a transition for existing kubeconfig consumers.

### DIFF
--- a/controllers/user/controllers/helper/kubeconfig/sa.go
+++ b/controllers/user/controllers/helper/kubeconfig/sa.go
@@ -131,7 +131,11 @@ func (sac *ServiceAccountConfig) applyBoundTokenSecret(
 		return nil, err
 	}
 	if secret.UID == "" {
-		return nil, fmt.Errorf("bound token secret %s/%s has empty uid", secret.Namespace, secret.Name)
+		return nil, fmt.Errorf(
+			"bound token secret %s/%s has empty uid",
+			secret.Namespace,
+			secret.Name,
+		)
 	}
 	return secret, nil
 }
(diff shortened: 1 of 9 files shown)


controllers/user/controllers/helper/kubeconfig/sa.go
```diff
@@ -61,93 +71,38 @@ func (sac *ServiceAccountConfig) applyServiceAccount(_ *rest.Config, client clie
...
-	secret := &v1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      sac.secretName,
-			Namespace: sac.namespace,
-		},
+	tokenRequest, err := clientset.CoreV1().
+		ServiceAccounts(sac.namespace).
+		CreateToken(ctx, sac.user, &authenticationv1.TokenRequest{
+			Spec: authenticationv1.TokenRequestSpec{
+				ExpirationSeconds: ptr.To(int64(sac.tokenRequestExpirationSeconds())),
```
zijiren233, line 190: Is `boundObjectRef` required to automatically invalidate the token?
For example, binding to a specific Secret?

controllers/user/controllers/helper/kubeconfig/sa.go
```diff
@@ -105,6 +142,10 @@ func (sac *ServiceAccountConfig) tokenRequestExpirationSeconds() int32 {
 	return sac.expirationSeconds
 }
 
+func TokenSecretName(name string) string {
```
zijiren233, line 219: I believe a random secret name should be used here, and the secret name should be stored before generation to ensure recoverability. When rotating keys, the old secret should be deleted, and the secret name should also be rotated.

zijiren233 (changes requested):

bxy4543 (approved):

zijiren233 (approved):

PR #7203, opened 2026-08-11
fix(user): make kube API client rate limit configurable

dinoallo:
## Background

User controller reconciles were hitting client-side throttling when listing and updating many users, which slows kubeconfig rotation and related reconciliation work.

## Changes

- Add `--kube-api-qps` and `--kube-api-burst` flags to the user controller.
- Thread those settings through the chart, generated deploy manifest, and entrypoint script.
- Expose `kubeAPI.qps` and `kubeAPI.burst` in the user controller values files.
- Update the default user controller deployment patch and README examples to match the new flags.
- Remove the old `config-file-path` wiring from the user controller deployment path.

## Risks / Notes

- Default client-side limits are now explicit and can be tuned per deployment.
- Deployments that relied on the mounted config file need the updated chart/entrypoint path.

### DIFF
--- a/controllers/user/deploy/user-controller-entrypoint.sh
+++ b/controllers/user/deploy/user-controller-entrypoint.sh
@@ -1,16 +1,42 @@
 #!/bin/bash
-set -euo pipefail
+set -eo pipefail
 
 HELM_OPTS=${HELM_OPTS:-""}
+HELM_OPTIONS=${HELM_OPTIONS:-""}
 RELEASE_NAME=${RELEASE_NAME:-"user-controller"}
 RELEASE_NAMESPACE=${RELEASE_NAMESPACE:-"user-system"}
 CHART_PATH=${CHART_PATH:-"./charts/user-controller"}
+HELM_SET_ARGS=()
+
+get_cm_value() {
+  local namespace="$1"
+  local name="$2"
+  local key="$3"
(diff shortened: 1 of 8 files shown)


controllers/user/deploy/user-controller-entrypoint.sh
```diff
@@ -1,16 +1,42 @@
...
+# Auto configure from sealos-config
+SEALOS_CLOUD_DOMAIN=${SEALOS_CLOUD_DOMAIN:-"$(get_cm_value sealos-system sealos-config cloudDomain)"}
+SEALOS_CLOUD_PORT=${SEALOS_CLOUD_PORT:-"$(get_cm_value sealos-system sealos-config cloudPort)"}
+
+if [ -n "${SEALOS_CLOUD_DOMAIN}" ]; then
+  add_set_string cloudAPIServerDomain "${SEALOS_CLOUD_DOMAIN}"
+fi
+
+if [ -n "${SEALOS_CLOUD_PORT}" ]; then
+  add_set_string cloudAPIServerPort "${SEALOS_CLOUD_PORT}"
```
zijiren233: Is this cloud port not the API server port?

zijiren233 (changes requested):

zijiren233 (approved):

PR #7208, opened 2026-08-12
feat(desktop): route Stripe returns to declared app

aimeritething:
Extends the workspace subscription Stripe payment flow so the client can declare which desktop app initiated the payment (`payApp`), and routes the Stripe success return back to that app (costcenter by default, Brain billing for `system-brain`).

- Add a `PayApp` enum, validate it on the pay endpoint, persist it on workspace subscription transactions (with a Cockroach migration for the global DB).
- Propagate `payApp` into the Stripe Checkout success/cancel URLs and include it in the same-pending-request check.
- Desktop resolves the `app` callback param and opens the declared app after a successful Stripe return.

**Known limitation:** the redirect only exists on the Checkout Session paths (new subscription / upgrade from Free). Upgrades between paid plans go through a Stripe hosted invoice (`HostedInvoiceURL`), which never redirects back to the desktop, so `payApp` has no effect on that path even though it is persisted.

The Brain frontend change that sends `payApp: "system-brain"` lives outside this repo and will follow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/frontend/desktop/src/utils/stripeCallback.ts
+++ b/frontend/desktop/src/utils/stripeCallback.ts
@@ -1,5 +1,4 @@
-const COSTCENTER_APP_KEY = 'system-costcenter';
-const BRAIN_APP_KEY = 'system-brain';
+import { BRAIN_APP_KEY, COSTCENTER_APP_KEY } from '@/constants/app';
 
 export const resolveStripeCallbackTarget = (payApp: string | string[] | undefined) => {
   if (payApp === BRAIN_APP_KEY) {
--- a/frontend/desktop/src/utils/stripeCallback.ts
+++ b/frontend/desktop/src/utils/stripeCallback.ts
@@ -1,0 +1,16 @@
+const COSTCENTER_APP_KEY = 'system-costcenter';
+const BRAIN_APP_KEY = 'system-brain';
+
+export const resolveStripeCallbackTarget = (payApp: string | string[] | undefined) => {
+  if (payApp === BRAIN_APP_KEY) {
+    return {
+      appKey: BRAIN_APP_KEY,
+      pathname: '/billing'
+    } as const;
+  }
+
+  return {
+    appKey: COSTCENTER_APP_KEY,
+    pathname: '/'
+  } as const;
+};
--- a/frontend/desktop/src/utils/stripeCallback.ts
+++ b/frontend/desktop/src/utils/stripeCallback.ts
@@ -1,7 +1,8 @@
 import { BRAIN_APP_KEY, COSTCENTER_APP_KEY } from '@/constants/app';
 
 export const resolveStripeCallbackTarget = (payApp: string | string[] | undefined) => {
-  if (payApp === BRAIN_APP_KEY) {
+  const normalizedPayApp = Array.isArray(payApp) ? payApp[0] : payApp;
+  if (normalizedPayApp === BRAIN_APP_KEY) {
     return {
       appKey: BRAIN_APP_KEY,
       pathname: '/billing'
--- a/service/account/api/workspace_subscription.go
+++ b/service/account/api/workspace_subscription.go
@@ -1389,7 +1389,6 @@ func CreateWorkspaceSubscriptionPay(c *gin.Context) {
 			if req.PromotionCode != "" {
 				transaction.StatusDesc = legacyPromotionCodeStatusPrefix + req.PromotionCode
 			}
-
 		}
 
 	case types.SubscriptionTransactionTypeDowngraded:
@@ -1486,11 +1485,7 @@ var ErrSamePendingOperation = errors.New("same pending operation exists")
 const legacyPromotionCodeStatusPrefix = "Promotion code: "
 
 func pendingTransactionPromotionCode(description string) string {
-	if !strings.Contains(description, legacyPromotionCodeStatusPrefix) {
-		return ""
-	}
-
-	parts := strings.Split(description, legacyPromotionCodeStatusPrefix)
+	parts := strings.SplitN(description, legacyPromotionCodeStatusPrefix, 2)
 	if len(parts) < 2 {
 		return ""
 	}
--- a/service/account/api/workspace_subscription.go
+++ b/service/account/api/workspace_subscription.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math"
@@ -1162,8 +1163,21 @@ func handleCalculatedUpgrade(
 // @Param req body WorkspaceSubscriptionOperatorReq true "WorkspaceSubscriptionOperatorReq"
 // @Success 200 {object} WorkspaceSubscriptionPayResp
 // @Router /payment/v1alpha1/workspace-subscription/pay [post]
-func CreateWorkspaceSubscriptionPay(c *gin.Context) {
+func parseWorkspaceSubscriptionPayReq(
+	c *gin.Context,
+) (*helper.WorkspaceSubscriptionOperatorReq, error) {
 	req, err := helper.ParseWorkspaceSubscriptionOperatorReq(c)
+	if err != nil {
+		return nil, err
+	}
+	if req.PayApp != "" && !req.PayApp.IsValid() {
+		return nil, fmt.Errorf("invalid payApp: %s", req.PayApp)
+	}
+	return req, nil
+}
+
+func CreateWorkspaceSubscriptionPay(c *gin.Context) {
+	req, err := parseWorkspaceSubscriptionPayReq(c)
 	if err != nil {
 		SetErrorResp(
 			c,
@@ -1263,6 +1277,16 @@ func CreateWorkspaceSubscriptionPay(c *gin.Context) {
 	//	req.Operator = types.SubscriptionTransactionTypeCreated
 	//}
 
+	requestStatusDescription, err := encodePendingWorkspaceSubscriptionRequestIdentity(req, false)
+	if err != nil {
+		SetErrorResp(
+			c,
+			http.StatusInternalServerError,
+			gin.H{"error": fmt.Sprintf("failed to encode payment request identity: %v", err)},
+		)
+		return
+	}
+
 	// Create subscription transaction with direct operator usage
 	transaction := types.WorkspaceSubscriptionTransaction{
 		ID:           uuid.New(),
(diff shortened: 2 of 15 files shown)


service/account/api/workspace_subscription_pay_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/labring/sealos/controllers/pkg/types"
+	"github.com/labring/sealos/service/account/helper"
+)
+
+func TestParseWorkspaceSubscriptionPayReqRejectsInvalidPayApp(t *testing.T) {
+	gin.SetMode(gin.TestMode)
+	recorder := httptest.NewRecorder()
+	ctx, _ := gin.CreateTestContext(recorder)
+	ctx.Request = httptest.NewRequest(
```
norberia: golangci-lint fails here: `httptest.NewRequest` → `httptest.NewRequestWithContext` (noctx, line 17) and `"POST"` → `http.MethodPost` (usestdlibvars, line 18).

service/account/helper/request_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/gin-gonic/gin"
+)
+
+func TestParseWorkspaceSubscriptionOperatorReqAcceptsAllowedPayApps(t *testing.T) {
+	gin.SetMode(gin.TestMode)
+	for _, payApp := range []string{"system-costcenter", "system-brain"} {
+		t.Run(payApp, func(t *testing.T) {
+			recorder := httptest.NewRecorder()
+			ctx, _ := gin.CreateTestContext(recorder)
+			ctx.Request = httptest.NewRequest(
```
norberia: Same lint failures as in `workspace_subscription_pay_test.go`, here and at lines 46–47: noctx (`httptest.NewRequestWithContext`) and usestdlibvars (`http.MethodPost`).

service/account/api/workspace_subscription.go
```diff
@@ -1372,10 +1386,10 @@ func CreateWorkspaceSubscriptionPay(c *gin.Context) {
 				transaction.Amount = 0
 			}
 
-			// Store promotion code for later use in Stripe session creation
 			if req.PromotionCode != "" {
-				transaction.StatusDesc = "Promotion code: " + req.PromotionCode
+				transaction.StatusDesc = legacyPromotionCodeStatusPrefix + req.PromotionCode
 			}
+
```
norberia: This added blank line is the `whitespace` lint failure (`unnecessary trailing newline`) — drop it.

service/pkg/pay/stripe.go
```diff
@@ -62,11 +62,17 @@ func getBaseURL() string {
 func buildURLs(
 	s *StripeService,
 	transaction *types.WorkspaceSubscriptionTransaction,
+	payApp types.PayApp,
 ) (string, string) {
 	// https://192.168.10.35.nip.io/?openapp=system-costcenter?payId%3D4W86BOp70ltT%26stripeState%3Dsuccess%26transactionId%3D3d09bc07-976c-44bc-8b52-05619693056e%26workspaceId%3Dns-47f3dbxj
 	baseURL := s.Domain + "/?payId=" + transaction.PayID + "&workspaceId=" + transaction.Workspace + "&transactionId=" + transaction.ID.String() + "&stripeState="
 	successURL := baseURL + "success"
 	cancelURL := baseURL + "cancel"
+	if payApp != "" {
```
norberia, line 71: Note: the desktop callback only routes on `stripeState === 'success'`, so the `app` param appended to the cancel URL is currently ignored (cancel returns go through the session-restore flow). Fine to keep for future use, just flagging it's dead weight today.

While touching this function: the example URL in the comment above (`?openapp=system-costcenter?payId%3D...`) predates the current format and is now doubly stale — worth updating.

service/account/api/workspace_subscription.go
```diff
@@ -1469,6 +1483,31 @@ func CreateWorkspaceSubscriptionPay(c *gin.Context) {
...
 var ErrSamePendingOperation = errors.New("same pending operation exists")
 
+const legacyPromotionCodeStatusPrefix = "Promotion code: "
+
+func pendingTransactionPromotionCode(description string) string {
+	if !strings.Contains(description, legacyPromotionCodeStatusPrefix) {
+		return ""
+	}
+
+	parts := strings.Split(description, legacyPromotionCodeStatusPrefix)
```
norberia: nit: `strings.SplitN(description, legacyPromotionCodeStatusPrefix, 2)` would make the `Contains` check and the `len(parts)` guard unnecessary. Behavior is preserved from the old inline code either way, and extracting it into a function is an improvement.

norberia (commented):
Reviewed the full flow (pay API → `pay_app` column → Stripe success/cancel URLs → desktop routing). The design is sound: the client declares `payApp` explicitly instead of the server guessing the origin, both the backend (`IsValid()`) and the desktop (`resolveStripeCallbackTarget`) whitelist the value, and including `PayApp` in the same-pending-request check is necessary — otherwise a brain-initiated request could reuse a pending session whose success URL still redirects to costcenter. Legacy pending transactions (empty `pay_app`) keep matching requests that omit the field (covered by a test), and the migration correctly targets only the global DB since Localdb doesn't have this table.

Two points beyond the inline comments:

1. **No caller sends `payApp` yet.** Neither costcenter nor anything else in this repo sets it, and the brain frontend lives outside this repo. Is the brain-side change (sending `payApp: "system-brain"` in the pay request) tracked somewhere? Until that lands this is unused plumbing.

2. **Paid-plan upgrades won't route back.** The redirect only exists on the Checkout Session paths (new subscription / upgrade from Free). Upgrades between paid plans go through a hosted invoice (`HostedInvoiceURL`), which never redirects back to the desktop, so `payApp` has no effect there even though it's persisted. Worth stating this limitation in the PR description.

Also a typo in the PR title: "decalred" → "declared" (it ends up in the release notes).

CI: `build-services (account) / golangci-lint` is failing with 7 issues, all in code added by this PR — see inline comments.

service/account/api/workspace_subscription_pay_test.go
aimeritething: Fixed in bfa611863 — switched to `httptest.NewRequestWithContext` and `http.MethodPost`.

service/account/helper/request_test.go
aimeritething: Fixed in bfa611863 — both occurrences now use `httptest.NewRequestWithContext` and `http.MethodPost`.

service/account/api/workspace_subscription.go
aimeritething: Dropped in bfa611863.

service/pkg/pay/stripe.go
aimeritething: Updated the stale example URL comment in bfa611863. Keeping the `app` param on the cancel URL intentionally — it keeps success/cancel construction symmetric and is ready if the cancel path ever routes by app; noted it is unused by the desktop today.

service/account/api/workspace_subscription.go
aimeritething: Applied in bfa611863 — now uses `strings.SplitN(..., 2)`; the `len` guard stays since `SplitN` returns a single element when the prefix is absent.

service/pkg/pay/stripe.go
```diff
@@ -62,11 +62,17 @@ func getBaseURL() string {
 func buildURLs(
 	s *StripeService,
 	transaction *types.WorkspaceSubscriptionTransaction,
+	payApp types.PayApp,
 ) (string, string) {
 	// https://192.168.10.35.nip.io/?openapp=system-costcenter?payId%3D4W86BOp70ltT%26stripeState%3Dsuccess%26transactionId%3D3d09bc07-976c-44bc-8b52-05619693056e%26workspaceId%3Dns-47f3dbxj
```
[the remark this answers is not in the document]
aimeritething: Updated in bfa611863 — the comment now shows the actual `buildURLs` output format.

service/account/helper/request_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"strings"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+)
+
+func TestParseWorkspaceSubscriptionOperatorReqAcceptsAllowedPayApps(t *testing.T) {
+	gin.SetMode(gin.TestMode)
+	for _, payApp := range []string{"system-costcenter", "system-brain"} {
+		t.Run(payApp, func(t *testing.T) {
```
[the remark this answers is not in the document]
aimeritething: Renamed to `TestParseWorkspaceSubscriptionOperatorReqBindsPayApp` in bfa611863. The arbitrary-value case is covered by `TestParseWorkspaceSubscriptionOperatorReqLeavesPayAppValidationToPayEndpoint`.

frontend/desktop/src/utils/stripeCallback.ts
```diff
@@ -0,0 +1,15 @@
...
+      appKey: BRAIN_APP_KEY,
+      pathname: '/billing'
+    } as const;
+  }
+
+  return {
+    appKey: COSTCENTER_APP_KEY,
+    pathname: '/'
+  } as const;
+};
```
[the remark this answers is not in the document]
aimeritething: Fixed in bfa611863 — array values are normalized to their first element, with a unit test covering the repeated-param case.

aimeritething:
Thanks for the thorough review — all inline comments addressed in bfa611863 (lint fixes, `string[]` normalization on the desktop resolver, `SplitN` simplification, test rename, refreshed URL comment).

On the two review-level points:

1. **Brain-side caller**: the Brain frontend change that sends `payApp: "system-brain"` is in development in the Brain repo and will follow this PR.
2. **Paid-plan upgrades**: correct — the hosted-invoice path never redirects back to the desktop. Added this as a known limitation in the PR description.

Also fixed the title typo (declared).

service/pkg/pay/stripe.go
```diff
@@ -62,11 +62,17 @@ func getBaseURL() string {
 func buildURLs(
 	s *StripeService,
 	transaction *types.WorkspaceSubscriptionTransaction,
+	payApp types.PayApp,
 ) (string, string) {
-	// https://192.168.10.35.nip.io/?openapp=system-costcenter?payId%3D4W86BOp70ltT%26stripeState%3Dsuccess%26transactionId%3D3d09bc07-976c-44bc-8b52-05619693056e%26workspaceId%3Dns-47f3dbxj
+	// Example: https://cloud.sealos.io/?payId=4W86BOp70ltT&workspaceId=ns-47f3dbxj&transactionId=3d09bc07-976c-44bc-8b52-05619693056e&stripeState=success&app=system-brain
```
dinoallo: Is this a leak?

zjy365 (approved):

zjy365 (approved):

bxy4543 (approved):

PR #7209, opened 2026-08-12
fix(desktop): expose pod workspace quota

zjy365:
## Summary

- map ResourceQuota `hard.pods` and `used.pods` to a public quota item with `type: "pod"`
- add pod quota support to Desktop, client SDK, shared types, formatting, icons, i18n, and quota guards
- cover the workspace quota API mapping with focused unit tests

## Why

User quota CRDs already contain pod limits, but the Desktop workspace quota API omitted them, so downstream consumers could not display or guard against pod limits.

## Validation

- Desktop workspace quota API tests: 2/2
- shared package TypeScript check
- targeted Prettier and ESLint
- `git diff --check`

Full workspace TypeScript compilation is unavailable in the sparse checkout because unrelated generated Prisma clients and package outputs are absent.

## Related change

The corresponding brain PR consumes and displays this new quota item. Its link will be added here after both PRs are created.

### DIFF
--- a/frontend/desktop/src/__tests__/unit/backend/workspace/getQuota.api.test.ts
+++ b/frontend/desktop/src/__tests__/unit/backend/workspace/getQuota.api.test.ts
@@ -1,0 +1,81 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { verifyAccessToken, generateBillingToken, fetchQuota } = vi.hoisted(() => ({
+  verifyAccessToken: vi.fn(),
+  generateBillingToken: vi.fn(() => 'billing-token'),
+  fetchQuota: vi.fn()
+}));
+
+vi.mock('@/services/backend/auth', () => ({
+  generateBillingToken,
+  verifyAccessToken
+}));
+
+import handler from '@/pages/api/workspace/getQuota';
+
+const createMockRes = () => {
+  const res: any = {
+    body: undefined,
+    json: vi.fn((payload: unknown) => {
+      res.body = payload;
+      return res;
+    })
+  };
+  return res;
+};
+
+describe('workspace quota api', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    (global as any).AppConfig = {
+      desktop: {
+        auth: {
+          billingUrl: 'https://billing.example.com'
+        }
+      }
+    };
+    vi.stubGlobal('fetch', fetchQuota);
+    verifyAccessToken.mockResolvedValue({
+      userId: 'user-id',
+      userUid: 'user-uid',
+      workspaceId: 'workspace-id'
+    });
+  });
+
+  it('maps the ResourceQuota pods entry to a pod quota item', async () => {
+    fetchQuota.mockResolvedValue({
+      clone: () => ({
+        json: async () => ({
+          quota: {
+            hard: { pods: '20' },
+            used: { pods: '3' }
+          }
+        })
+      })
+    });
+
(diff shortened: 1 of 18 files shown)


zjy365:
Consumer integration: https://github.com/labring/brain/pull/263

That PR consumes this new `type: "pod"` workspace quota item and displays pod usage and limits in the brain sidebar.

norberia:
Reviewed together with labring/brain#263. The direction is right and I verified the data path end to end: `GetWorkspaceResourceQuota` (service/account/api/workspace.go) passes through the ResourceQuota `hard`/`used` as-is, and `DefaultResourceQuotaHard()` (controllers/pkg/resources/resources.go) already includes `pods` (default 20, also configurable per subscription plan). So the premise "the data is already there, only the API omits it" holds.

**Blocking: CI is red — dbprovider and template image builds (amd64/arm64, 4 jobs) fail to compile.** From the dbprovider build log:

```
./src/components/QuotaBox/index.tsx:42:23
Type error: ... can't be used to index type '{ cpu...; nodeport...; traffic... }'.
  Property 'pod' does not exist ...
> 42 | const color = sourceMap[type]?.color;
```

Root cause: this PR adds `pod` to the shared `WorkspaceQuotaItemType` and makes the API return the pod item unconditionally, but three provider apps render `useUserQuota()` items directly and were not adapted:

| Consumer | Issue |
|---|---|
| `providers/dbprovider/src/components/QuotaBox/index.tsx` | `sourceMap` has no `pod` entry → **compile error**; also zh `t('pod')` resolves to the existing "实例" (DB-instance context), inconsistent with quota-dialog's "Pod" |
| `providers/template/src/pages/deploy/components/QuotaBox.tsx` | same `sourceMap` indexing → **compile error**; locales have no `pod` key at all, label renders as raw "pod" |
| `providers/applaunchpad/src/pages/app/edit/components/QuotaBox.tsx` | `iconMap` is `Record<string, ...>` so it compiles, but the pod row has **no icon**; top-level `t('pod')` key missing (existing one is nested under the logs namespace) |

Needed: add `pod` color/icon entries to the three QuotaBoxes, add the missing i18n keys for applaunchpad/template, and get image-build green.

Non-blocking notes:

- `formatResourceQuotaValue` uses `toFixed(2)`, so pod tooltips show "20.00". nodeport already has this issue — fine to leave, but worth a follow-up to format count-typed resources as integers.
- The same type union is hand-written in 6 places (desktop zod, shared zod, client-sdk d.ts, two hooks, store). All updated here, but this structure guarantees the next addition misses one. Suggest a follow-up deriving the requirements types from `WorkspaceQuotaItemTypeSchema`.
- Test is solid (mock boundary, full response assertions, matches vitest include). Minor: `vi.stubGlobal('fetch', ...)` has no `unstubAllGlobals` cleanup.
- costcenter has a fully independent quota path (`providers/costcenter/src/pages/api/workspace/get-workspace-quota.ts` + own types) that doesn't expose pod. Nothing breaks (its zod schema never sees the new value), but its valuation page won't show pods — suggest a follow-up issue rather than expanding this PR.
- `client-sdk` type change needs an npm release of `@labring/sealos-desktop-sdk`, otherwise the brain-side local shim can never be removed. Worth listing as a follow-up here.

Cross-repo ordering: merge/deploy this PR before labring/brain#263 — the brain guard safely filters unknown items in both directions, but if brain ships first users see a "Pods --/--" row until Desktop upgrades.

Nixieboluo (approved):

PR #7217, opened 2026-08-13
fix(desktop): preserve marketing attribution through Cloud auth

yangchuansheng:
## Summary

- preserve bounded `sea_attr` and `consent_token` through Cloud auth, workspace, and region handoffs
- keep `login_success` at authoritative login-completion paths only (OAuth callback, email, and phone); remove page-load and handoff emissions that caused duplicate or false events
- issue a user-bound, short-lived Brain receipt bound to the raw attribution hash
- fail closed for consent: browser-provided consent fields are transport data only and are always signed as `unspecified` until a trusted CMP / Google Consent Mode source is wired
- clear persisted attribution on logout, auth expiry, and account deletion

## Validation

- `pnpm exec vitest run --project unit src/__tests__/unit/backend/auth/marketing-consent.test.ts src/__tests__/unit/utils/marketing-attribution.test.ts src/__tests__/unit/utils/gtm.test.ts` — 12 passed
- `pnpm exec tsc --noEmit`
- Prettier and `git diff --check`

## Consent contract

`sea_attr` is not evidence of consent. The signed receipt proves the authenticated subject and attribution payload hash, while consent remains `unspecified` until a trusted consent source is integrated. Brain must continue redacting click IDs and enhanced-conversion data for this state.

## Release notes

Brain must be deployed with the compatible marketing contract and the same signing value as Cloud `jwtMarketingConsent` / Brain `MARKETING_CONSENT_SIGNING_KEY`.

Companion Brain PR: https://github.com/labring/brain/pull/267

### DIFF
--- a/frontend/desktop/data/config.yaml
+++ b/frontend/desktop/data/config.yaml
@@ -95,6 +95,7 @@ desktop:
       internal: "thisisinternaljwt"
       regional: "thisisregionaljwt"
       global: "thisisglobaljwt"
+      marketingConsent: ""
     billingUrl: "http://account-service.account-system.svc:2333"
     workorderUrl: ""
     cloudVitrualMachineUrl: ""
--- a/frontend/desktop/deploy/HELM_VALUES_GUIDE.md
+++ b/frontend/desktop/deploy/HELM_VALUES_GUIDE.md
@@ -58,6 +58,7 @@ desktopConfig:
   jwtInternal: 'your-jwt-internal-key' # 内部 JWT 密钥
   jwtRegional: 'your-jwt-regional-key' # 区域 JWT 密钥
   jwtGlobal: 'your-jwt-global-key' # 全局 JWT 密钥
+  jwtMarketingConsent: 'your-marketing-consent-key' # Marketing consent JWT secret shared with Brain
 ```
 
 ### 4. 计费配置
--- a/frontend/desktop/deploy/HELM_VALUES_GUIDE_CN.md
+++ b/frontend/desktop/deploy/HELM_VALUES_GUIDE_CN.md
@@ -58,6 +58,7 @@ desktopConfig:
   jwtInternal: 'your-jwt-internal-key' # 内部 JWT 密钥
   jwtRegional: 'your-jwt-regional-key' # 区域 JWT 密钥
   jwtGlobal: 'your-jwt-global-key' # 全局 JWT 密钥
+  jwtMarketingConsent: 'your-marketing-consent-key' # 与 Brain 共享的营销 consent JWT 密钥
 ```
 
 ### 4. 计费配置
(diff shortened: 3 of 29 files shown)


zjy365 (commented):
**P1 — `login_success` fires twice per login in the template-deploy flow**

For OAuth logins, `login_success` is already emitted in `/callback` (`callback.tsx`, right after `sessionConfig`, with `oauth2_provider` and the correct `user_type`). The Brain template-deploy flow — ad → `/oauth?templateName=…` → OAuth → `/callback` → redirect back to `/oauth` — now emits the event **twice for one login**:

1. `/callback`: `user_type: 'new'`, includes `oauth2_provider`
2. `/oauth`: `user_type: 'existing'` (since `switchRegionType` is not `INIT`), **no** `oauth2_provider`

Three problems:

- **Double counting**: one login produces two `login_success` events.
- **Contradictory attributes**: `new` vs `existing`, and `oauth2_provider` is missing on the `/oauth` emission.
- **False positives**: `/oauth` fires the event on every load with a valid session (re-opening a deploy link, cross-region redirect), not only on actual logins — this inflates the funnel data this event feeds.

Suggestion: emit `login_success` in exactly one place that owns the login completion — `/callback` already does. Either remove the call from `/oauth`, or, if the cross-region `INIT` flow (which does not pass through `/callback`) still needs it, scope it to `switchRegionType === INIT` **and** add a session-scoped dedupe guard (e.g., record `userUid` after the first emission) so repeat visits cannot re-fire it.

Same pattern applies to the new emissions in `Workspace.tsx` / `switchRegion.tsx` — worth unifying into one helper with the dedupe guard.

zjy365 (commented):
**P2 — Consent state is client-asserted, not server-verified**

In `consent-token.ts`, `browserConsentState()` decodes the client-supplied `sea_attr` and copies `ad_personalization` / `ad_user_data_consent` into the signed token. Any logged-in user can craft a version 2/3 `sea_attr` with `ad_user_data_consent: "granted"` and obtain a fully valid token — the signature proves the token is untampered, but the underlying consent claim is self-reported by the browser.

This feeds Google Ads (offline/enhanced conversions), so it is a compliance risk, not just data quality. Also, `consent_source` is hardcoded to `"desktop_oauth"`, so downstream (Brain) cannot distinguish an asserted consent from one verified by a CMP / Consent Mode.

This limitation is acknowledged in the design doc, but it is not stated in the PR body or release notes.

Suggestion (non-blocking): document the risk explicitly in the PR description / release notes, and plan to derive consent from a real CMP / Google Consent Mode source instead of the URL parameter (keeping `consent_source` as the discriminator for downstream policy).

yangchuansheng:
Addressed zjy365 feedback in c01fb64c6. login_success now remains only on authoritative OAuth callback, email, and phone completion paths; OAuth/workspace/region handoff pages no longer emit it. The consent receipt still binds the user and attribution hash, but browser-asserted consent is fail-closed as unspecified until a trusted CMP or Google Consent Mode source is integrated. Updated tests and PR contract notes.

zjy365 (approved):

zijiren233 (approved):

PR #7230, opened 2026-08-17
fix(account): prevent stale debt deletion after recharge

dinoallo:
## Background

Recharge recovery could race with debt-driven namespace deletion. A stale replay or scheduled deletion path could continue using an earlier debt state and overwrite a namespace that had already resumed.

## Changes

- Coordinate debt refresh, replay, and payment processing per user.
- Revalidate the current debt state before replaying final deletion.
- Preserve resume state when handling replay annotations.
- Revalidate `FinalDeletion` immediately before destructive namespace cleanup.
- Cancel stale deletion reconciliation instead of marking a recovered namespace as `FinalDeletionCompleted`.

## Risks / Notes

Kubernetes deletion requests that have already been accepted cannot be cancelled. This change prevents the account controller from sending stale deletion requests and stops destructive cleanup when the namespace has recovered before execution.

### DIFF
--- a/controllers/account/controllers/namespace_controller.go
+++ b/controllers/account/controllers/namespace_controller.go
@@ -74,6 +74,8 @@ const (
 	DeployPVCResizeKey = "deploy.cloud.sealos.io/resize"
 )
 
+var errFinalDeletionCancelled = errors2.New("final deletion cancelled by namespace status")
+
 //+kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=namespaces/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=core,resources=namespaces/finalizers,verbs=update
@@ -146,6 +148,9 @@ func (r *NamespaceReconciler) Reconcile(
 	// auxiliary function handles resource operations
 	performAction := func(action func(context.Context, string) error, actionName string) (ctrl.Result, error) {
 		if err := action(ctx, req.Name); err != nil {
+			if errors2.Is(err, errFinalDeletionCancelled) {
+				return ctrl.Result{}, err
+			}
 			logger.Error(err, actionName+" namespace resources failed")
 			return ctrl.Result{
 				Requeue:      actionName == deleteConst,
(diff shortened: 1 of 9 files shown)


controllers/account/controllers/namespace_controller.go
```diff
@@ -472,11 +514,19 @@ func (r *NamespaceReconciler) DeleteUserResource(ctx context.Context, namespace
...
 		}(rs)
 	}
+	wasCancelled := false
 	for range deleteResources {
 		if err := <-errChan; err != nil {
+			if errors2.Is(err, errFinalDeletionCancelled) {
+				wasCancelled = true
+				continue
+			}
 			return err
```
zijiren233, line 544: Is this check redundant? Can we directly retain the original logic `return err`?

controllers/account/controllers/namespace_controller.go
```diff
@@ -2849,13 +2915,21 @@ func deleteResourceListAndWait(
 		close(errCh)
 	}()
 
+	wasCancelled := false
 	for deleteErr := range errCh {
+		if errors2.Is(deleteErr, errFinalDeletionCancelled) {
+			wasCancelled = true
+			continue
```
zijiren233, line 2946: It’s similar here; an error should be returned directly.

controllers/account/controllers/namespace_controller.go
```diff
@@ -441,10 +450,43 @@ func (r *NamespaceReconciler) deleteBackup(ctx context.Context, namespace string
...
+		gvr,
+		namespace,
+		r.deleteBackupSemaphore,
+		func(ctx context.Context) error {
+			return r.ensureFinalDeletionActive(ctx, namespace)
+		},
+	)
+}
+
+func (r *NamespaceReconciler) ensureFinalDeletionActive(ctx context.Context, namespace string) error {
```
zijiren233: Design the return value as `bool, error` instead of using `error` alone as the return value.

zijiren233 (dismissed):

controllers/account/controllers/namespace_controller.go
dinoallo: I kept this check intentionally. DeleteUserResource starts multiple deletion workers. Returning immediately on errFinalDeletionCancelled could let sibling workers continue running after the reconcile has returned. A recharge/resume reconciliation could then run concurrently with in-flight deletion requests, reintroducing the race this PR is intended to fix.

For the cancellation path, we therefore keep collecting the worker results and return the cancellation signal only after the workers have finished. The outer reconcile treats this sentinel as an expected cancellation.

bxy4543 (approved):

PR #7237, opened 2026-08-19
fix(costcenter): embed Alipay cashier page via iframe

zjy365:
## Summary

The Alipay payment URL generated by `TradePagePay` is a signed gateway URL over 900 characters long, which cannot be rendered as a scannable QR code (QR version 30+, module size below the readable threshold).

This PR switches Alipay to `qr_pay_mode=4` (order code mode) so **Alipay itself renders the official cashier page** (including a clearly scannable QR code), returns the auto-submitting cashier form HTML as the payment code, and embeds it in the frontend via an iframe — the same approach FastGPT uses for Alipay payments.

## Changes

- `controllers/pkg/pay/alipay_payment.go`: `CreatePayment` uses `TradePagePay` with `QRPayMode=4` and `QRCodeWidth=210`, returning the cashier page HTML instead of the long signed URL.
- `frontend/providers/costcenter/src/components/RechargeModal.tsx`: `AlipayPayment` detects the HTML payment code and renders it in an `<iframe srcDoc>` (215×215); falls back to QR code rendering for legacy data.

## Notes

- WeChat payments are unaffected (they use the short native `weixin://` URL).
- Payment status polling via `TradeQuery` (`GetPaymentDetails`) is unchanged.
- Existing pending orders with the old long URL keep the QR fallback until they expire (12 min).

### DIFF
--- a/controllers/pkg/pay/alipay_payment.go
+++ b/controllers/pkg/pay/alipay_payment.go
@@ -3,6 +3,7 @@ package pay
 import (
 	"context"
 	"fmt"
+	"html"
 	"os"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


dinoallo (approved):
LGTM

Nixieboluo (approved):

PR #7240, opened 2026-08-20
fix(user): avoid full User objects in webhook validation

dinoallo:
## Background

Creating a User invokes the validating webhook. The webhook previously listed full User objects to calculate the user limit. User status can contain kubeconfig data, so the list response becomes very large and can exceed the Kubernetes admission webhook timeout.

## Changes

- Use the Kubernetes metadata client to count Users without fetching spec or status.
- Apply metadata-only counting in the User validating webhook and User reconciler license-limit check.
- Maintain the quota user count from User informer Add/Update/Delete events instead of manually listing all Users at controller startup.
- Wait for the registered informer event handler to sync before marking the counter initialized, ensuring initial Add events are applied before quota checks start.
- Remove the controller startup full User list used only to initialize the legacy count cache.
- Set the User validating webhook timeout to 30 seconds as a short-term fallback.
- Add unit coverage for metadata-based counting, event-driven counter updates, and webhook test setup.

## Risks / Notes

The quota counting behavior remains the same: deleting Users are excluded, and reconciliation excludes the current User. The counter is maintained by the existing informer event stream; no second full User cache or startup list is introduced. The timeout increase is only a fallback; the main fix is avoiding full User object retrieval.

## Verification

- `go test ./pkg/usercount`
- `go build -buildvcs=false ./...`
- `helm lint controllers/user/deploy/charts/user-controller`
- Helm template rendering confirms `vuser.kb.io` uses `timeoutSeconds: 30`.
- Envtest suites could not start in the restricted execution environment because local listener creation was denied.

### DIFF
--- a/controllers/user/controllers/user_count.go
+++ b/controllers/user/controllers/user_count.go
@@ -28,7 +28,7 @@ import (
 
 type userCountRunnable struct {
 	counter    *usercount.Counter
-	initialize func(context.Context) ([]interface{}, error)
+	initialize func(context.Context) ([]any, error)
 }
 
 func (r *userCountRunnable) Start(ctx context.Context) error {
@@ -59,7 +59,7 @@ func SetupUserCount(mgr ctrl.Manager) (*usercount.Counter, error) {
 		return nil, fmt.Errorf("add user count event handler: %w", err)
 	}
 
-	initialize := func(ctx context.Context) ([]interface{}, error) {
+	initialize := func(ctx context.Context) ([]any, error) {
 		if storeInformer, ok := informer.(interface{ GetStore() toolscache.Store }); ok {
 			return storeInformer.GetStore().List(), nil
 		}
@@ -68,7 +68,7 @@ func SetupUserCount(mgr ctrl.Manager) (*usercount.Counter, error) {
 		if err := mgr.GetClient().List(ctx, list); err != nil {
 			return nil, err
 		}
-		objects := make([]interface{}, len(list.Items))
+		objects := make([]any, len(list.Items))
 		for i := range list.Items {
 			objects[i] = &list.Items[i]
 		}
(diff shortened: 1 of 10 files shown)


zijiren233 (changes requested):
It appears that the user controller already caches the complete User CR. Should we streamline the cached fields and switch the metadata list operation to use the User CRs already cached by the controller, rather than using the metadata client to fetch the full set from the API server every time?

dinoallo:
> It appears that the user controller already caches the complete User CR. Should we streamline the cached fields and switch the metadata list operation to use the User CRs already cached by the controller, rather than using the metadata client to fetch the full set from the API server every time?

This is a valid concern. The manager already maintains a structured informer cache for full User objects, but using it for every quota check would require listing and deep-copying the complete objects, including status.kubeConfig. The current metadata client only retrieves PartialObjectMetadata, so it avoids the large full-object payload while still incurring an API-server LIST per admission request. For this PR we kept the metadata-only path; a follow-up should consider maintaining a local count from informer events or using a dedicated metadata informer cache to avoid repeated API-server lists.

controllers/user/controllers/user_count.go
```diff
@@ -16,15 +16,75 @@ package controllers
...
+		list := &userv1.UserList{}
+		if err := mgr.GetClient().List(ctx, list); err != nil {
+			return nil, err
+		}
+		objects := make([]any, len(list.Items))
+		for i := range list.Items {
+			objects[i] = &list.Items[i]
+		}
+		return objects, nil
+	}
```
zijiren233: When the informer starts, it processes existing resources as "add" events; therefore, manual initialization is neither required nor advisable, as it could lead to data consistency issues caused by resource version mismatches.

zijiren233 (changes requested):

zijiren233 (approved):

PR #7253, opened 2026-08-24
fix(subscription): allow deleted plans to resubscribe

zjy365:
## Summary

- treat `DELETED` subscriptions as recreatable in the account service instead of rejecting them as active
- route expired CostCenter subscriptions to the resubscribe flow with the `created` operator
- keep canceled subscriptions resumable only before their current period expires
- add backend and frontend regression coverage for deleted, debt, and normal states

## Testing

- `go test ./api -run "TestCanRecreateWorkspaceSubscription|TestDeletedWorkspaceSubscriptionIsExpired|TestParseWorkspaceSubscriptionPayReqRejectsInvalidPayApp|TestSamePendingWorkspaceSubscriptionRequestIncludesPayApp" -count=1`
- `pnpm --filter costcenter test:ci` (23 tests passed)
- `pnpm --filter costcenter lint` (passed with existing warnings)

## Local baseline limitations

- full `go test ./api` requires configured database clients and currently panics in the existing credits tests when they are absent
- `pnpm --filter costcenter exec tsc --noEmit` is blocked by the existing mock typing error in `refetch-after-payment.test.ts`

### DIFF
--- a/frontend/providers/costcenter/__tests__/unit/subscription-state.test.ts
+++ b/frontend/providers/costcenter/__tests__/unit/subscription-state.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, it } from 'vitest';
+import {
+  canRecreateSubscription,
+  getSubscriptionOperator,
+  isSubscriptionExpired
+} from '@/utils/subscription';
+
+const currentPlan = {
+  UpgradePlanList: ['Pro'],
+  DowngradePlanList: ['Free']
+};
+
+describe('subscription state', () => {
+  it.each(['DEBT', 'deleted'])('allows %s subscriptions to be recreated', (status) => {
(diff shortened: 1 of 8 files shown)


dinoallo:
We should not treat every DELETED workspace as resubscription-ready. DELETED does not indicate whether the namespace and its resources are still recoverable: some workspaces may already be terminating or in final deletion, while others are only canceled at the subscription level. Allowing all of them could create a NORMAL database record for an unrecoverable namespace, cause orphaned resources or double billing, and allow late Stripe events from the old subscription to revive the workspace incorrectly.
The safer behavior is to allow resubscription only when the namespace is still active and recoverable, create a new Stripe subscription, and validate the transaction and webhook subscription ID before restoring the workspace.

dinoallo:
https://github.com/labring/sealos/pull/7258

PR #7254, opened 2026-08-25
perf(account): optimize workspace consumption aggregation

dinoallo:
## Background

The `/account/v1alpha1/costs/workspace/consumption` endpoint calculates settled consumption grouped by namespace. The previous pipeline used `$facet` to aggregate both `app_costs` and top-level `amount`, so the two branches reprocessed the same matched Billing documents and could double-count a record when it exposed both the authoritative top-level total and nested resource breakdown.

## Changes

- Add the common `owner`, `status`, `time`, and optional `namespace` filters before aggregation. No `type` filter is applied because settled LLM token records are stored as `SubConsumption` and must remain included alongside `Consumption` records.
- Use a single `$match` + `$group` pipeline over top-level `amount` for the normal workspace-total request, avoiding `$facet` and `$unwind` entirely.
- Only inspect `app_costs` when `appName` filtering is requested, using `$filter`/`$map`/`$sum` to keep one output row per Billing record.
- Make direct `APP-STORE`/`LLM-TOKEN` matching mutually exclusive from nested resource matching to prevent duplicate totals.
- Normalize and validate `appType`, handle cursor decoding errors, and add pipeline unit tests for the default and app-filtered paths.
- Add the `{owner: 1, status: 1, time: 1}` billing index for the common query path.

## Risks / Notes

`Billing.Amount` is treated as the authoritative total for a Billing record, consistent with the existing billing writers. App-level filtering still sums matching `app_costs.amount` for nested resource types and uses top-level `amount` once for direct application types. The API response shape is unchanged.

Validation completed locally:

- `cd service && go test ./account/dao -run 'Test(NormalizeWorkspaceConsumptionAppType|BuildWorkspaceConsumptionPipeline)' -count=1`
- `cd service && go test ./account/dao -run '^$' -count=1`
- `golangci-lint run --config=.golangci.yml` (from the repository root)
- `cd service/account && go test ./dao -run '^TestGetWorkspaceConsumptionAmountWithMongoRuntime$' -count=1 -v` (testcontainers with MongoDB `4.4.29`; 5 scenarios passed)
- `cd service/account && go test ./dao -run '^$' -bench '^BenchmarkGetWorkspaceConsumptionAmountWithMongoRuntime$' -benchtime=100ms -count=1 -benchmem`

  The benchmark uses 10,000 billing records and excludes container startup and fixture insertion from the timed section. The local reference results were:

  | Case | Result |
  | --- | ---: |
  | `all` | 16.7 ms/op, 28,916 B/op, 407 allocs/op |
  | `namespace` | 5.6 ms/op, 11,269 B/op, 133 allocs/op |
  | `app_type` | 19.2 ms/op, 29,008 B/op, 408 allocs/op |
  | `app_name` | 34.8 ms/op, 39,730 B/op, 512 allocs/op |

  These local benchmark values are for comparison and are not used as a CI pass/fail threshold.

## Benchmark

The endpoint was benchmarked against a Kubernetes deployment with 50,000 temporary settled `APP` billing records across 100 namespaces. Each record contained 50 `app_costs` entries. The same request was sent directly to account-service five times for each image, using a 90-second client timeout; the frontend's 40-second timeout was not used for measurement.

| Deployment | Image digest | Result |
| --- | --- | --- |
| Baseline | `sha256:b0f432707d718cdf2b5fa19e462ea37eb6da8a9b7f6a5652eb0d628a88cdc60e` | 1.397-1.603s, median 1.408s |
| Optimized | `sha256:8c754d77513e996c09ec04c59c60e1e5b42abb769213e2853c2af93ee120939c` | 0.142-0.172s, median 0.158s |

All 10 requests returned HTTP 200 with the expected result: 100 namespaces, 50,000 per namespace, and a 3,112-byte response. Average latency decreased from 1.445s to 0.157s, an 89% reduction and approximately 9.2x speedup. The benchmark cluster did not yet have the new `{owner: 1, status: 1, time: 1}` index, so these numbers measure the aggregation pipeline optimization without the index contribution. The temporary billing records were deleted after the test and the original deployment image was restored.

Build: [GitHub Actions run](https://github.com/dinoallo/sealos/actions/runs/33495282704) (lint, amd64/arm64 image builds, image release, and cluster image build all passed).

### DIFF
--- a/service/account/dao/workspace_consumption_test.go
+++ b/service/account/dao/workspace_consumption_test.go
@@ -1,0 +1,127 @@
+package dao
+
+import (
+	"testing"
+	"time"
+
+	"github.com/labring/sealos/controllers/pkg/resources"
+	"github.com/labring/sealos/service/account/helper"
+	"go.mongodb.org/mongo-driver/bson"
+)
+
+func workspaceConsumptionRequest() helper.ConsumptionRecordReq {
+	return helper.ConsumptionRecordReq{
+		TimeRange: helper.TimeRange{
+			StartTime: time.Date(2026, time.January, 1, 0, 0, 0, 0, time.UTC),
+			EndTime:   time.Date(2026, time.January, 2, 0, 0, 0, 0, time.UTC),
+		},
+		Namespace: "ns-test",
+		AuthBase: helper.AuthBase{
+			Auth: &helper.Auth{Owner: "owner-test"},
+		},
+	}
+}
+
+func workspaceConsumptionStageValue(stage bson.D, key string) (any, bool) {
+	for _, element := range stage {
+		if element.Key == key {
+			return element.Value, true
+		}
+	}
+	return nil, false
+}
+
+func TestNormalizeWorkspaceConsumptionAppType(t *testing.T) {
+	normalized, value, err := normalizeWorkspaceConsumptionAppType(" app-store ")
+	if err != nil {
+		t.Fatalf("normalize app type: %v", err)
+	}
+	if normalized != resources.AppStore {
+		t.Fatalf("normalized app type = %q, want %q", normalized, resources.AppStore)
+	}
+	if value != resources.AppType[resources.AppStore] {
+		t.Fatalf("app type value = %d, want %d", value, resources.AppType[resources.AppStore])
+	}
+
+	if _, _, err := normalizeWorkspaceConsumptionAppType("unknown"); err == nil {
+		t.Fatal("expected an error for an unsupported app type")
+	}
+}
+
+func TestBuildWorkspaceConsumptionPipelineWithoutAppFilter(t *testing.T) {
+	pipeline, err := buildWorkspaceConsumptionPipeline(workspaceConsumptionRequest())
+	if err != nil {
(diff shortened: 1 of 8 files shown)


service/account/dao/workspace_consumption_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/labring/sealos/controllers/pkg/resources"
+	"github.com/labring/sealos/service/account/helper"
+	"go.mongodb.org/mongo-driver/bson"
+)
+
```
zijiren233, line 11: I believe partial performance tests should be added, using Testcontainers to simulate real data and verify whether execution time has returned to an acceptable level, along with gating mechanisms to prevent future performance regressions.

bxy4543 (approved):

service/account/dao/workspace_consumption_test.go
dinoallo: Tests have been added in another file

zijiren233 (approved):

PR #7257, opened 2026-08-25
perf(resources): reduce controller cache memory

zijiren233:
## What this PR does

- project cached Pods, PVCs, Backups, Services, Namespaces, Instances, and the GPU ConfigMap to fields used by the resource controller
- route monitor reads through the projected cache while keeping write paths on complete API server objects
- clear managed fields and fail fast when an informer was not explicitly registered
- add projection tests for every cached resource type

## Verification

- `go test ./...`
- envtest
- `go build ./...`
- `go vet ./...`
- `golangci-lint`
- `git diff --check`

### DIFF
--- a/controllers/resources/controllers/cache/cache.go
+++ b/controllers/resources/controllers/cache/cache.go
@@ -1,0 +1,358 @@
+// Copyright 2026 labring.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cache
+
+import (
+	"context"
+	"fmt"
+
+	kbv1alpha1 "github.com/apecloud/kubeblocks/apis/dataprotection/v1alpha1"
+	appv1 "github.com/labring/sealos/controllers/app/api/v1"
+	"github.com/labring/sealos/controllers/pkg/gpu"
+	"github.com/labring/sealos/controllers/pkg/resources"
+	accounttypes "github.com/labring/sealos/controllers/pkg/types"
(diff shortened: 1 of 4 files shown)


zijiren233:
53-cluster verification completed:

- Build Controller workflow passed: golangci-lint, amd64/arm64 controller image, and cluster image.
- Tested `ghcr.io/zijiren233/sealos-resources-controller@sha256:b08cc9d552e8cf90aff9ed209ba0eaa7ad456651e7efa0cdc0da6f3cb9affcaf`; rollout reached 1/1 Ready with 0 restarts.
- MonitorReconciler completed seven consecutive one-minute cycles. Two cycles included a temporary owner-labeled namespace, a Running Pod with CPU/memory/ephemeral-storage requests and limits, and a NodePort Service.
- Cluster inventory during the test included 72 PVCs, 139 Backups, 12 Instances, and 168 Services. No missing-informer, field-index, type-conversion, list, monitor, database-insert, panic, or fatal errors were observed.
- Stable controller memory was 19 MiB versus the 54 MiB baseline, approximately 65% lower.
- The existing missing `node-system/node-gpu-info` ConfigMap error remained unchanged from the baseline.
- NetworkReconciler and NamespaceQuotaReconciler are already disabled in `main.go`, so they were outside the active runtime test scope.
- The temporary namespace was deleted and the original image/digest restored. Final inventory returned to 222 Pods, 167 Services, and 64 Namespaces.

zijiren233:
Review finding follow-up:

- Confirmed the startup race: with `ReaderFailOnMissingInformer=true`, controller-runtime 0.20.4 uses `Peek` and does not wait for `HasSynced`; the two constructor retries could finish before the ConfigMap informer synchronized, leaving GPU maps empty for the first billing interval.
- Fixed in `799d9863e` by waiting for all explicitly registered informers to synchronize before constructing MonitorReconciler.
- Local verification passed: full tests, targeted race tests, build, vet, golangci-lint, and diff check.
- Rebuild workflow passed for lint, amd64/arm64 image, and cluster image: https://github.com/zijiren233/sealos/actions/runs/32884525192
- Targeted 53-cluster test used a temporary GPU ConfigMap whose node name matched no real node. The very first startup refresh reported `aliasCount=1` and `nodeCount=1` without retry or cache errors, followed by two successful billing cycles. The tested pod stayed Ready with 0 restarts and 18 MiB memory.
- Temporary resources were deleted and the original controller image/digest was restored.

bxy4543 (approved):

PR #7268, opened 2026-08-27
fix(user): enforce namespace security policy

zijiren233:
## Summary
- make the admin namespace Pod Security labels and legacy cluster-admin binding configurable, disabled by default
- watch metadata-only Namespace and ClusterRoleBinding changes and restore User ownership metadata
- apply Pod Security labels to all ns-* namespaces by default, including namespaces without a Sealos User
- keep orphan namespace labels unchanged when strict mode is disabled
- reduce cached metadata and propagate startup migration errors

## Configuration
- `adminClusterAdmin.enabled`: preserve legacy admin cluster-admin behavior when enabled
- `strictNamespacePodSecurity.enabled`: default `true`; when disabled, orphan ns-* namespaces are left unchanged

## Tests
- `go test ./controllers/cache`
- targeted User controller tests
- `go test -race ./controllers`
- `go vet ./controllers/...`
- `helm lint deploy/charts/user-controller`
- `helm template user-controller deploy/charts/user-controller`
- `git diff --check`


cuisongliu (approved):

zijiren233:
Superseded by refactor/user-controller; opening a replacement PR with the consolidated refactor scope.

PR #7293, opened 2026-09-03
fix(account): expose startup and readiness probes

dinoallo:
## Background

The account controller can take a long time to initialize database clients, controllers, and background processors. Its health endpoints are currently registered only when the manager starts, so Kubernetes cannot distinguish slow startup from an unhealthy process and may repeatedly restart the pod.

## Changes

- Serve `/healthz`, `/startupz`, and `/readyz` from a standalone probe server before expensive initialization begins.
- Keep liveness available while startup is in progress.
- Report startup success after dependency and controller registration completes.
- Report readiness after the controller-runtime manager has started its cache and leader runnables.
- Configure the deployment manifests to use `/startupz` for the startup probe.
- Add endpoint state transition tests.

## Risks / Notes

The probe server binds the configured health-probe address, so the controller-runtime health probe listener is disabled to avoid a port conflict.

### DIFF
--- a/controllers/account/main.go
+++ b/controllers/account/main.go
@@ -424,6 +424,9 @@ func main() {
 	//+kubebuilder:scaffold:builder
 
 	if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+		if !mgr.GetCache().WaitForCacheSync(ctx) {
+			return nil
+		}
 		probeState.markReady()
 		<-ctx.Done()
 		return nil
--- a/controllers/account/main.go
+++ b/controllers/account/main.go
@@ -46,8 +46,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlconfig "sigs.k8s.io/controller-runtime/pkg/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
-	"sigs.k8s.io/controller-runtime/pkg/healthz"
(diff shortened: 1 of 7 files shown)


controllers/account/main.go
```diff
@@ -415,12 +423,15 @@ func main() {
 
 	//+kubebuilder:scaffold:builder
 
-	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
-		setupLog.Error(err, "unable to set up health check")
-		os.Exit(1)
-	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
-		setupLog.Error(err, "unable to set up ready check")
+	if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
```
zijiren233: `manager.RunnableFunc` does not implement `LeaderElectionRunnable`, so `mgr.Add` places it into the leader-election runnable group. During a single-replica rolling update, the old Pod holds the lease, preventing the new Pod from executing `markReady()`, thus continuously returning HTTP 503; meanwhile, the default `maxUnavailable=0` retains the old Pod, causing a deployment deadlock. Please ensure this runnable implements `NeedLeaderElection() == false`, and mark it ready after cache synchronization completes on each replica.

zijiren233 (changes requested):
The `make build` error may need to be fixed.

zijiren233 (approved):
LGTM

PR #7294, opened 2026-09-03
perf(account): optimize consumption amount aggregation

dinoallo:
## Background

`/account/v1alpha1/costs/consumption` used `$facet` as the first aggregation stage. This prevented the common owner, status, and time filters from using the billing index efficiently and expanded nested cost arrays with `$unwind`.

## Changes

- Move the common filters into the first `$match` stage.
- Replace the two `$facet` branches with per-document `$reduce` and `$filter` calculations followed by one `$group`.
- Preserve the existing nested and direct consumption amount semantics.
- Add MongoDB Testcontainers coverage for nested costs, direct AppStore/LLMToken costs, namespaces, and app filters.
- Add a 10,000-record MongoDB benchmark for the optimized query.
- Run both consumption runtime tests by default in the account CI job.

## Risks / Notes

- The existing behavior for app-name-only requests is intentionally preserved for compatibility.
- The benchmark uses MongoDB 4.4.29 through Testcontainers and is available for performance verification; it is not used as a CI pass/fail threshold.

## Verification

- `gofmt` passed.
- `git diff --check` passed.
- Local runtime execution was not completed because the environment could not download the required Go 1.25 toolchain; the runtime tests are configured as a required account CI job.

### DIFF
--- a/.github/workflows/service-build.yml
+++ b/.github/workflows/service-build.yml
@@ -87,7 +87,7 @@ jobs:
         working-directory: service/account
         env:
           TESTCONTAINERS_REQUIRED: "true"
-        run: go test ./dao -run '^TestGetWorkspaceConsumptionAmountWithMongoRuntime$' -count=1 -v
+        run: go test ./dao -run '^TestGet(Workspace)?ConsumptionAmountWithMongoRuntime$' -count=1 -v
 
   image-build:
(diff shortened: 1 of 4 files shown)


zijiren233 (approved):
LGTM

PR #7302, opened 2026-09-04
perf(account): add startup stage timing logs

dinoallo:
## Background

Account controller startup can spend a long time between initialization messages, making it difficult to identify which startup stage is responsible for delayed readiness.

## Changes

- Add elapsed-time logs for MongoDB, CockroachDB, configuration loading, controller setup, billing/index initialization, and health checks.
- Measure manager cache synchronization independently so slow informer synchronization is visible.
- Report completed stages with `duration_ms` while keeping startup behavior and health endpoints unchanged.

## Risks / Notes

- Observability-only change; no business behavior is intended to change.
- Timing logs contain stage names and durations only; they do not add credentials or configuration payloads.

### DIFF
--- a/controllers/account/main.go
+++ b/controllers/account/main.go
@@ -84,14 +84,12 @@ func beginStartupStage(name string) func(error) {
 	return func(err error) {
 		duration := time.Since(startedAt)
 		if err != nil {
-			setupLog.Info(
+			setupLog.Error(err,
 				"startup stage failed",
 				"stage",
 				name,
 				"duration_ms",
 				duration.Milliseconds(),
-				"error",
-				err,
 			)
 			return
 		}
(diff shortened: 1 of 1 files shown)


bxy4543 (approved):

zijiren233 (approved):

controllers/account/main.go
```diff
@@ -77,6 +78,40 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
+func beginStartupStage(name string) func(error) {
+	startedAt := time.Now()
+	setupLog.Info("startup stage started", "stage", name)
+	return func(err error) {
+		duration := time.Since(startedAt)
+		if err != nil {
+			setupLog.Info(
```
zijiren233: A higher log level should be used.

PR #7330, opened 2026-09-15
fix(applaunchpad): restore HPA spec verbatim on resume

zjy365:
## What

Fixes two bugs in the applaunchpad pause/resume flow around HPA:

1. **CPU target jumps from 50% to 500% (×10 applied twice)**: pause saved the live HPA's `averageUtilization` (already scaled ×10), and start fed that value back into `json2HPA`, multiplying it by 10 again.
2. **Elastic policy breaks after a scheduled or manual start**: on resume `storeList` was missing, so a Deployment got an HPA pointing at a StatefulSet; and the startup replica count used the request's literal value instead of the saved `minReplicas`.

## How

Add `src/utils/pauseResume.ts` and route all three start/stop paths (`pauseApp`/`startApp`, `updateReplica`'s `PauseApp`/`StartApp`, and the replica branches of `updateAppResources`) through it:

- **Pause**: capture the full live HPA `spec` into the pause annotation (keeping `target`/`value` for display).
- **Resume**: when a `spec` exists, `yaml.dump` it verbatim — never going through `json2HPA` again, which removes both the double ×10 and the workload-kind guess. For legacy annotations without a `spec`, divide the value by 10 and derive `storeList` from the live `app.kind`.
- **Elastic startup replicas**: take the saved `spec.minReplicas` instead of the requested `+replica`.
- **Repeat-pause guard**: an existing saved `spec` is not overwritten with an empty payload.

## Tests

- All 169 existing unit tests pass.
- Manual round-trip check: CPU 50% → HPA 500 → pause saves spec → start restores HPA 500 (page shows 50%); `scaleTargetRef.kind` correct for both Deployment and StatefulSet; GPU HPA metric name and absolute target value preserved; repeat pause does not clobber restore data.


Nixieboluo (approved):
lgtm

